PR #130, opened 2014-01-15
Allow use of an external class loader to resolve class names when there are split packages

PeterC-DLS:
We have a problem where the Class.forName calls used in Py4J fail owing to our use of packages that are split across different plugins.

To solve this, I've added code to use a class loader in place of that call. A OSGI service interface is defined and if a preference is set then any implementation is used to retrieve a class loader; if there is no such service, then the normal call is used. On our side, we implement a class loader using a CompositeClassLoader to pull together several class loaders that span the relevant plugins.


jonahkichwacoders:
Peter, because this pull request was done against dawn's master branch, new commits unrelated to the pull request are ending up in the pull request.

Issue #133, opened 2014-04-04
py4j hangs if java methods use System.out.println


davidcsterratt:
I've been using System.out.println() statements in java objects. However, when I run these objects via py4j 0.8.1 in python2.7, there are two effects:
1. None of the standard output appears to come through to the python terminal (I'm running via plain python2.7 -i)
2. After a while the process seems to hang; CPU activity goes down to zero, but the call to the Java method doesn't return anything. 
I can prevent the hang happening by removing System.out.println() statements from the java code. 

Is this a known issue? I've not seen anything in the documentation or via google searches about it. If the issue isn't known about, I can try to give a small reproducible example.

 I'm guessing that there is some buffer in py4j that is being filled up and not flushing.

bartdag:
After a long long time (apologies!), I think I have an elegant solution (still need some time to nail down the details). I believe that the output of the Java program might be interesting for some users, so discarding the output after the initial port exchange might not be always desirable.

Furthermore, temp files could grow very large for long running processes.

launch_gateway will thus accept a parameter: output_redirect. 

Once the port exchange is done, a daemon thread will be started and will read the output and send it to output_redirect:
* If output_redirect is None, it will write the input to os.devnull (portable version of /dev/null). This is the most memory and space efficient solution.
* If output_redirect is a file descriptor, it will write to the file
* If output_redirect is a Queue or a deque, it will put/appendleft (very useful in threading context)

Some small details:
* Need to determine what to do with stderr (could be two daemon threads... small optimization would be to send stderr to os.devnull if the caller is not interested in the output of stderr).

I did not want to start a thread, but since there is no way to change the stdout/stderr of the subprocess once it is started, there is little choice. This solution is both cross-platform and only involves the python part.

Again, thanks for providing code example, it's been very helpful..

davidcsterratt:
Thanks for getting back onto this and no worries about the delay. I'm not much of a python expert, but what you suggest sounds very sensible.

bartdag:
The fix has been merged in the @master branch. Feel free to reopen the ticket if that still does not work.

PR #138, opened 2014-06-18
Potential fix for #133, by redirecting pipe to stdout after port read

davidcsterratt:
I've confirmed that this prevents the deadlock in the sample code I posted in #133. However, this does seem to be a hack - it would be nice to a way of doing this that didn't need another subprocess...


davidcsterratt:
Apologies for reviving this, but I'm trying to work out if this issue has now been resolved in some other way - there didn't seem to be a resolution to it, but I've not used py4j recently, nor checked through the code.

bartdag:
Hi, yes, the issue was resolved starting from this line:
https://github.com/bartdag/py4j/blob/master/py4j-python/src/py4j/java_gateway.py#L173

I ended up creating threads to consume the output and prevent a deadlock. There are now various options to discard the output or consume it.

PR #154, opened 2015-01-07
Allow passing additional arguments to launched JavaGateways when created via launch_gateway

Julian:


py4j-python/src/py4j/java_gateway.py
```diff
@@ -1182,12 +1184,14 @@ def launch_gateway(cls, port=0, jarpath="", classpath="", javaopts=[],
...
+        Additional keyword arguments are passed along to the newly created
+        instance.
+
         :rtype: a :class:`JavaGateway <py4j.java_gateway.JavaGateway>`
             connected to the `Gateway` server.
         """
         _port = launch_gateway(port, jarpath, classpath, javaopts, die_on_exit)
-        gateway = JavaGateway(GatewayClient(port=_port))
-        return gateway
+        return cls(GatewayClient(port=_port), **kwargs)
```
bartdag: nice catch (cls instead of JavaGateway).

The main issue I see is that GatewayClient is deprecated and should be passed in an instance of GatewayParameters. Maybe instead of kwargs, the method should access an instance of GatewayParameters and CallbackServerParameters (these could be extended to pass extra parameters to a subclass of Gateway).

bartdag:
Thanks for your pull request and for the tests! As stated in my commit comment, my only issue is with the use of GatewayClient. This parameter will be deprecated in the next version and the way to fix it would clash with **kwargs. I would suggest to use the new GatewayParameters and CallbackServerParameters instead.

Julian:
Hey, thanks for the quick response. I'll have a look at the other object and see if I see what you're suggesting.

Julian:
OK, I think I might see what you're trying to do with those two objects (replace the parameters that you were directly giving to `JavaGateway` by two objects containing those parameters -- I'm not sure I see directly why yet, but that appears to be what those parameter objects are doing?)

In which case it'd seem like `kwargs` is compatible with that, if you want to pass those parameters in, you can do that -- the way it's incompatible is in that it always creates a `GatewayClient` instead of using the newer way for creating one, but that's true even without that patch. Is that what you mean?  You'd like `launch_gateway` to be reimplemented to use the new way?

bartdag:
> I'm not sure I see directly why yet, but that appears to be what those parameter objects are doing?

I get many requests to add parameters or new ways to configure a JavaGateway or a CallbackServer so instead of having n parameters in the JavaGateway constructor, I'm grouping these parameters in two objects. These objects could be extended to provide even more parameters in the future for custom gateways.

> the way it's incompatible is in that it always creates a GatewayClient instead of using the newer way for creating one, 

Exactly

> but that's true even without this patch

You are right. Your pull request made me realize this :-)

>  You'd like launch_gateway to be reimplemented to use the new way?

I see a few strategies:

1. We keep the port parameter, but instead of passing it to a GatewayClient, we wrap it in a GatewayParameters object. The problem is that if kwargs contain a GatewayParameters, we must use the GatewayParameters instance provided in kwargs to set the port. A bit messy.
2. We keep the port parameter and the GatewayClient, but we add a deprecation warning. If someone wants to pass a port, they need to pass it in a GatewayParameters instance in kwargs. Once we will remove the port parameter, the signature won't be very helpful/self-documenting though.
3. We add gateway_parameters and callback_server_parameters to launch_gateway and we deprecate the port parameter. While port still exists, we wrap it in a GatewayParameters instance (the one provided or a new one if is None). With gateway_parameters, I'm not sure kwargs is really helpful, but we could keep it.

PR #156, opened 2015-03-05
copied all py4j sources to Eclipse

jonahkichwacoders:
copied all py4j sources to Eclipse


jonahkichwacoders:
Don't know why this pull request failed to build, only the Java8 sub-builds failed. This is part of pull 157 anyway, so can be closed if 157 is taken

bartdag:
I've been fighting with timeouts for a long time, but it is difficult to wrap a client and a server in the same test client and mocking everything will not help detect most of the protocol errors I've seen. Do not worry about timeouts, especially if it works locally and you did not change the code that is being tested.

Thanks for your pull requests. I'll definitively take a look and comment on them soonish (it might have to wait until next weekend).

PR #157, opened 2015-03-05
Interactive interpretter improvements

jonahkichwacoders:
This set of commits includes a number of changes to improve interoperability of Py4J with interactive environments. In particular PyDev and IPython.

At the top level, this means that __dir__ and __doc__ have been implemented for the java_gateway classes.

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -827,7 +827,7 @@ class JavaObject(object):
         return 'JavaObject id=' + self._target_id
 
 
-class JavaClass():
+class JavaClass(object):
     """A `JavaClass` represents a Java Class from which static members can be
        retrieved. `JavaClass` instances are also needed to initialize an array.
 
@@ -961,7 +961,7 @@ class UserHelpAutoCompletion(object):
     KEY = "<package or class name>"
 
 
-class JavaPackage():
+class JavaPackage(object):
     """A `JavaPackage` represents part of a Java package from which Java
        classes can be accessed.
 
@@ -982,6 +982,9 @@ class JavaPackage():
     def __getattr__(self, name):
         if name == UserHelpAutoCompletion.KEY:
             return UserHelpAutoCompletion
+
+        if name in ['__str__', '__repr__']:
+            raise AttributeError
 
         if name == '__call__':
             raise Py4JError('Trying to call a package.')
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -976,7 +976,13 @@ class JavaPackage():
             self._jvm_id = DEFAULT_JVM_ID
         self._jvm_id = jvm_id
 
+    def __dir__(self):
+        return [UserHelpAutoCompletion.KEY]
+
     def __getattr__(self, name):
+        if name == UserHelpAutoCompletion.KEY:
+            return UserHelpAutoCompletion
+
         if name == '__call__':
             raise Py4JError('Trying to call a package.')
         new_fqn = self._fqn + '.' + name
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -765,6 +765,8 @@ class JavaObject(object):
         return list(set(self._methods.keys()) | self._field_names)
 
     def _populate_fields(self):
+        # Theoretically, not thread safe, but the worst case scenario is
+        # cache miss or double overwrite of the same method...
         if not self._fully_populated:
             if self._auto_field:
                 command = DIR_COMMAND_NAME +\
@@ -854,6 +856,8 @@ class JavaClass(object):
 
 
     def __dir__(self):
+        # Theoretically, not thread safe, but the worst case scenario is
+        # cache miss or double overwrite of the same method...
         if self._statics is None:
             command = DIR_COMMAND_NAME +\
                 DIR_STATIC_SUBCOMMAND_NAME +\
@@ -1024,14 +1028,13 @@ class JVMView(object):
             # for regular Py4J classes.
             self._jvm_object = jvm_object
 
-        self._dir_sequence = None
-        self._dir_cache = []
+        self._dir_sequence_and_cache = None, []
 
     def __dir__(self):
         command = DIR_COMMAND_NAME +\
             DIR_JVMVIEW_SUBCOMMAND_NAME +\
             self._id + '\n' +\
-            get_command_part(self._dir_sequence) + '\n' +\
+            get_command_part(self._dir_sequence_and_cache[0]) + '\n' +\
             END_COMMAND_PART
 
         answer = self._gateway_client.send_command(command)
@@ -1039,9 +1042,10 @@ class JVMView(object):
                 self._fqn, "__dir__")
         if return_value is not None:
             result = return_value.split('\n')
-            self._dir_sequence = result[0]
-            self._dir_cache = result[1:] + [UserHelpAutoCompletion.KEY]
-        return self._dir_cache[:]
+            # Theoretically, not thread safe, but the worst case scenario is
+            # cache miss or double overwrite of the same method...
+            self._dir_sequence_and_cache = result[0], result[1:] + [UserHelpAutoCompletion.KEY]
+        return self._dir_sequence_and_cache[1][:]
 
     def __getattr__(self, name):
         if name == UserHelpAutoCompletion.KEY:
--- a/py4j-python/src/py4j/tests/java_dir_test.py
+++ b/py4j-python/src/py4j/tests/java_dir_test.py
@@ -153,3 +153,10 @@ def test_dir_jvmview_two():
 
             eq_(sorted(dir(view1)), [UserHelpAutoCompletion.KEY, "Class1", "Class2", "Class3", "Class4"])
             eq_(sorted(dir(view2)), [UserHelpAutoCompletion.KEY, "Class1", "Class2", "Class3", "Class5"])
+
+def test_dir_package():
+    with example_app_process():
+        with gateway() as g:
+            eq_(sorted(dir(g.jvm)), [UserHelpAutoCompletion.KEY])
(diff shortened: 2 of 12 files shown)


jonahkichwacoders:
I am reviewing the errors re other version of Python (I guess you can tell I wrote in 2.7)

jonahkichwacoders:
You can see a youtube video of how DawnScience (http://www.dawnsci.org/) are using py4j here https://www.youtube.com/watch?v=zhmLMMqPI6Q

What this new code will do (when fully integrated) is support easier use for users of Dawn in the interactive console.

bartdag:
Apologies for the delay in merging this pull request, my schedule has been more hectic than usual. I'll try to take a look as soon as possible (might be one or two weeks).

jonahkichwacoders:
Hopefully you feel a little less hectic soon, we are using a fork of py4j at Dawn anyway, so this is not on our critical path.

jonahkichwacoders:
Hi, I just wanted to give you a friendly ping on this pull request. I would really like to see it integrated. Is there anything I can do to make that easier?

py4j-python/src/py4j/tests/java_dir_test.py
```diff
@@ -0,0 +1,162 @@
...
+                      ])
+
+ExampleClassStatics = sorted([
+                              "StaticClass",
+                              "static_field",
+                              "static_method"
+                              ])
+
+@contextmanager
+def example_app_process():
```
bartdag: thanks, much cleaner this way :-)

py4j-python/src/py4j/tests/java_dir_test.py
```diff
@@ -0,0 +1,162 @@
+from __future__ import unicode_literals, absolute_import
+
+
+from py4j.java_gateway import JavaGateway, GatewayParameters, java_import,\
+    UserHelpAutoCompletion
+from py4j.tests.java_gateway_test import (
+    start_example_app_process)
+from contextlib import contextmanager
+from nose.tools import eq_
```
bartdag: let's stick to self.assertEqual or copy the eq_ code from nose: I try to keep Py4J dependency-free (even though the recommended way to run the test suite is with nose, this is not strictly required).

py4j-python/src/py4j/tests/java_dir_test.py
```diff
@@ -0,0 +1,162 @@
...
+
+@contextmanager
+def gateway(*args, **kwargs):
+    g = JavaGateway(
+        gateway_parameters=GatewayParameters(*args, auto_convert=True, **kwargs))
+    lineSep = g.jvm.System.lineSeparator()
+    try:
+        yield g
+        # Call a dummy method to make sure we haven't corrupted the streams
+        eq_(lineSep, g.jvm.System.lineSeparator())
```
bartdag: why did you need to do that?

py4j-python/src/py4j/tests/java_dir_test.py
```diff
@@ -0,0 +1,162 @@
...
+    with example_app_process():
+        with gateway() as g:
+            ex = g.getNewExample()
+            eq_(sorted(dir(ex)), ExampleClassMethods)
+
+def test_dir_object_fields():
+    with example_app_process():
+        with gateway(auto_field=True) as g:
+            ex = g.getNewExample()
+            eq_(sorted(dir(ex)), sorted(ExampleClassMethods + ExampleClassFields))
```
bartdag: very nice

py4j-python/src/py4j/tests/java_dir_test.py
jonahkichwacoders: > why did you need to do that? (I'm curious, nothing against it)
Because it was possible to write to the stream extraneous characters or insufficient characters. I wrote this after I realized I was testing a bunch of stuff, but I hadn't put the final separator at the end of the command.

py4j-python/src/py4j/tests/java_dir_test.py
```diff
@@ -0,0 +1,162 @@
...
+
+def test_dir_object_shows_manually_called_after_dir():
+    with example_app_process():
+        with gateway() as g:
+            ex = g.getNewExample()
+            eq_(sorted(dir(ex)), ExampleClassMethods)
+            try:
+                ex.does_not_exist_in_example()
+                raise AssertionError("Method should not have succeeded")
+            except:
```
bartdag: I would put the exact exception type or at least ``except Exception`` to avoid catching things like Ctrl-C (see https://www.python.org/dev/peps/pep-0008/#programming-recommendations "A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions")

py4j-python/src/py4j/tests/java_dir_test.py
```diff
@@ -0,0 +1,162 @@
...
+    with example_app_process():
+        with gateway() as g:
+            ex = g.getNewExample()
+            try:
+                ex.does_not_exist_in_example()
+                raise AssertionError("Method should not have succeeded")
+            except:
+                pass
+            # Make sure the manually called method now shows up
+            eq_(sorted(dir(ex)), sorted(ExampleClassMethods + ['does_not_exist_in_example']))
```
bartdag: From a usability perspective, why is this desirable (again, I'm more curious than opposed to the idea)?

py4j-python/src/py4j/java_gateway.py
```diff
@@ -794,7 +936,32 @@ def __call__(self, *args):
         return return_value
 
 
-class JavaPackage():
+class UserHelpAutoCompletion(object):
```
bartdag: I'm not 100% sure what the purpose of this class is. Is this a pattern/common use case for autocompletion in IPython or ?

py4j-python/src/py4j/java_gateway.py
```diff
@@ -678,6 +760,39 @@ def __getattr__(self, name):
         # The name is a method
         return self._methods[name]
 
+    def __dir__(self):
+        self._populate_fields()
+        return list(set(self._methods.keys()) | self._field_names)
+
+    def _populate_fields(self):
+        if not self._fully_populated:
```
bartdag: I would add a note similar to the one in __getattr__ stating that this is not thread safe but the worst case scenario is that this method is called twice (at the same time) and the _methods dict is filled twice with the same values. We may want to make everything thread-safe (or thread-efficient :-) ) later and this comment will be very helpful.

py4j-python/src/py4j/tests/java_dir_test.py
jonahkichwacoders: OK, I'll update for that. Actually I have started using py.test for most stuff and its assert rewriting is really nice, can't believe I wasn't using it earlier. So I'll udpate the code to use assert directly (can't use self.assertEquals because this code is not within a testcase class)

jonahkichwacoders: PS, In case you haven't come across it, here is a link https://pytest.org/latest/assert.html#assertfeedback

py4j-python/src/py4j/tests/java_dir_test.py
jonahkichwacoders: OK, update coming for this.

bartdag:
This looks really good. I asked a few questions and I will continue my review tomorrow (I just need to do some more tests and I want to make sure I fully understand the whole id/versioning of jvm views).

Thank you very much for this amazing pull request. The code is clear, there are tests and there are improvements I'll be able to make to other parts of Py4J code (e.g., the context manager in the tests).

py4j-python/src/py4j/tests/java_dir_test.py
bartdag: Thanks for the reference, this is indeed really nice and I'll probably add it to our company projects!

py4j-python/src/py4j/tests/java_dir_test.py
jonahkichwacoders: Good question. I am not sure of the right answer, I don't think it is great for usability, but the __dir__ in this case is reflecting the internal state of the JavaObject. The JavaObject is storing in self._methods the JavaMember regardless of whether the call succeeds. What could potentially be done is that once self._populate_fields() has been called error early on calling methods not already in self._methods. However I resisted erroring early in the first instance as I was concerned that I didn't affect existing behaviour of Py4J. But once that member is added to self._methods I didn't want to then ignore it. 

Perhaps the test would be more accurate as something like:

```python
            assert (sorted(dir(ex)) == sorted(ExampleClassMethods + ['does_not_exist_in_example'])) or (sorted(dir(ex)) == sorted(ExampleClassMethods)
```

As that would more accurately represent the contract I would like to see maintained going forward.

py4j-python/src/py4j/java_gateway.py
jonahkichwacoders: I am very much open to suggestions on the best way to deal with this. 

The use case I am working towards is users who don't really know what to type where, they certainly won't have read Py4J manual. So in places where it is not possible to provide completions, I wanted to give a clue to the user as to what they could type next. Unfortunately I can't say that this is a pattern or common use case in IPython or PyDev.

py4j-python/src/py4j/java_gateway.py
jonahkichwacoders: OK, I will add notes to the places I have.

Now that I am reviewing the code though I am concerned about the ```__dir__``` on the JVMView and what the effect of self._dir_sequence and self._dir_cache being out of step.

In the _current_ use case, ```__dir__``` it is always called from the same thread. But it is possible that someone complicates life in the future, so I'll have a think about it.

jonahkichwacoders: I am going to change the code to be a single tuple holding both the sequence and cache so that I don't have to worry about them getting out of step.

jonahkichwacoders:
I have added my updates based on your feedback. 

Please feel free to ask me to rebase and/or squash my changes if you would like.

bartdag:
I made a few small changes and merged your pull request. The support IPython can now provide is extremely satisfying :-)

Again, thanks for your contribution, it is greatly appreciated and it will be highlighted in the next release notes.

Issue #147, opened 2014-10-07
Use random port for Python callback server


davies:
We could use `0` as random port in Python Gateway, but can not get the real port which callback server was binded, and update the Java callback client.

243826:
I feel that this is similar to what I have been exploring. Let me explain my use  case:

1. Java Process Starts.
2. Java GatewayServer starts listening on port 0.
3. Communicates the actual port on which it's listening to Python Process which connects to this port and additionally starts a callbackserver on port 0.
4. Communicates the actual callbackserver port using entry point to JVM
5. JVM registers that as a port for CallbackClient

I think I am able to do 1-4. 5th although trivial to implement - I was wondering if there is a better workflow than what I am thinking to make both the processes to interoperate.

Please comment. I can work on the PR if you feel I am on the right track.

bartdag:
I guess in py4j.java_gateway.JavaGateway, we could add a method start_dynamic_callback_server(), which would:

1. Start the callback server on port 0
2. Communicate the port to the Java side (with a new command, because the GatewayServer instance is not accessible?)
3. Once receiving the port, the Java side would create a new CallbackClient with the port.

Would that be acceptable @davies ?

@243826 I am not sure how you would accomplish step 3: would the JVM starts the Python process?

davies:
@bartdag lgtm

243826:
@bartdag  After I was done implementing something on my end - I reread what @davies wanted. Realized that I wanted opposite of what he wanted. The main difference is that in my case JVM tells another agent to launch Python. This way I avoid parent/child relationship between the 2 processes. 

On the python side - I made the following change:
1. Provision to listen on port 0 was always there. I pass the callback server port explicitly as zero in my application. So no change there.
2. I added methods to query back the listening_port just the way Java api provides one.

I have not worked on special command to communicate the port since in my case I know the GatewayServer instance. Should I send in the PR for review?

bartdag:
Yes, I would like to see this. Methods that allow better introspection are definitively welcome.

Instead of a new command, I might just add a way to target the GatewayServer (e.g., gateway.java_server) which would allow greater customization.

Thanks!

bartdag:
@davies this is the solution @243826 contributed (I merely merged it and I added a java_gateway_server field to JavaGateway). Let me know if it works for you.

davies:
@bartdag @243826 That works for me, thanks! Will upgrade to 0.9.1 once it's released.

PR #179, opened 2015-12-08
Allow binding on the un predetermined port for the callback server

243826:
As discussed in #147 -  and do late initialization of the callback client.

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -699,7 +699,7 @@ class GatewayConnection(object):
             self.stream = self.socket.makefile("rb", 0)
         except Exception as e:
             msg = "An error occurred while trying to connect to the Java "\
-                "server"
+                "server ({0}:{1})".format(self.address, self.port)
             logger.exception(msg)
             raise Py4JNetworkError(msg, e)
 
@@ -1326,6 +1326,9 @@ class JavaGateway(object):
             self.shutdown()
             raise
 
+    def get_callback_server(self):
+        return self._callback_server
+
     def start_callback_server(self, callback_server_parameters=None):
         """Starts the callback server.
 
@@ -1598,8 +1601,11 @@ class CallbackServer(object):
             socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         try:
             self.server_socket.bind((self.address, self.port))
+            self._listening_address, self._listening_port = \
+                self.server_socket.getsockname()
         except Exception as e:
-            msg = "An error occurred while trying to start the callback server"
+            msg = "An error occurred while trying to start the callback "\
+                  "server ({0}:{1})".format(self.address, self.port)
             logger.exception(msg)
             raise Py4JNetworkError(msg, e)
 
@@ -1609,6 +1615,12 @@ class CallbackServer(object):
         # Default is False
         self.thread.daemon = self.callback_server_parameters.daemonize
         self.thread.start()
+
+    def get_listening_port(self):
+        return self._listening_port
+
+    def get_listening_address(self):
+        return self._listening_address
 
     def run(self):
         """Starts listening and accepting connection requests.
--- a/py4j-java/ant.properties
+++ b/py4j-java/ant.properties
@@ -27,7 +27,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 ###############################################################################
 # Version of the software
-version=0.9
+version=0.9.1-SNAPSHOT
 
 # Path to the JUnit 4 jar. YOU MUST SET THIS VALUE. Otherwise, the test classes won't compile.
 junit.path=junit-4.10.jar
--- a/py4j-java/pom.xml
+++ b/py4j-java/pom.xml
@@ -11,7 +11,7 @@
 	<artifactId>py4j</artifactId>
 	<packaging>jar</packaging>
 	<name>Py4J</name>
-	<version>0.9</version>
+	<version>0.9.1-SNAPSHOT</version>
 	<description>Py4J enables Python programs running in a Python interpreter to dynamically access Java objects in a Java Virtual Machine. Methods are called as if the Java objects resided in the Python interpreter and Java collections can be accessed through standard Python collection methods. Py4J also enables Java programs to call back Python objects.</description>
 	<url>http://www.py4j.org/</url>
 	<licenses>
--- a/py4j-java/src/py4j/CallbackClient.java
+++ b/py4j-java/src/py4j/CallbackClient.java
@@ -54,13 +54,9 @@ import java.util.logging.Logger;
 public class CallbackClient {
 	public final static String DEFAULT_ADDRESS = "127.0.0.1";
 
-	/**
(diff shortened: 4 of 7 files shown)


243826:
@bartdag attention please :-)

py4j-python/src/py4j/java_gateway.py
```diff
@@ -699,7 +699,7 @@ def start(self):
             self.stream = self.socket.makefile("rb", 0)
         except Exception as e:
             msg = "An error occurred while trying to connect to the Java "\
-                "server"
+                "server (%s:%s)" % (self.address, self.port)
```
bartdag: we use the .format version of string formatting :-)

py4j-python/src/py4j/java_gateway.py
```diff
@@ -1598,8 +1601,9 @@ def start(self):
             socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         try:
             self.server_socket.bind((self.address, self.port))
+            self.__listening_address, self.__listening_port = self.server_socket.getsockname()
         except Exception as e:
-            msg = "An error occurred while trying to start the callback server"
+            msg = "An error occurred while trying to start the callback server (%s:%d)" % (self.address, self.port)
```
bartdag: this is a very good idea. I should print these info as often as possible. Please use .format though.

py4j-python/src/py4j/java_gateway.py
```diff
@@ -1598,8 +1601,9 @@ def start(self):
             socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         try:
             self.server_socket.bind((self.address, self.port))
+            self.__listening_address, self.__listening_port = self.server_socket.getsockname()
```
bartdag: only one underscore (no magic attribute or name mangling), e.g., ``_listening_address``

bartdag:
Thanks for your pull request. This provides greater flexibility and visibility and I like it.

I need to think a little about the CallbackClient class.

1. I like that you can specify the address and the port of the CallbackClient after the GatewayServer is created. It makes sense.

2. I don't like the mutability of GatewayClient because I don't want the port and address to change after call back connections are created. I think this should be handled in GatewayServer, something like setupNewGatewayClient(address, port), which would a) stop and cleanup the existing gateway client and b) creates a new gateway client with immutable port and address. Not sure about the setupNewGatewayClient though.

What do you think?

243826:
Thanks for your review and appreciate it. I'll make the changes related to format and mangling and submit the pull request as you suggested.

I agree with you that setupNewGatewayClient as it seems cleaner as it clearly communicates the intention and gives an opportunity to do a cleanup. I would call it changePythonServerAddress(iface, port) or changePythonAddress(iface, port). But anyOtherName is good as well if you feel strongly about it.

Rationale:
May be it's just me that it took some time for me to wrap my head around clients/server concept  in py4j and how they map to Python side and JVM side. In python process - I expected GatewayServer would be something that python process launches and from JVM process jvm process would do the same. But the GatewayServer and CallbackServer meanings do not change and as a new comer I had to learn it by experimenting with ports. It's not that big of an issue but to make it super clear, I would have called them JVMServer and PythonServer instead of GatewayServer and CallbackServer respectively.

Let me know what your thoughts are and I may be able to code it up and update the pull request.

bartdag:
> I would have called them JVMServer and PythonServer instead of GatewayServer and CallbackServer respectively.

This is mainly due to evolution. At first, only one direction was supported (Java to Python) so this distinction was not required. I like the names you propose and may reuse them for the new threading model I will be working on next year.

243826:
I saw that my pull request failed the CI. I did not think that I changed anything to cause it. Will investigate and resubmit it. Thanks.

243826:
@bartdag I am trying to get to the root of the problems with the tests failing w/ travis CI and noticed that although the py4j-java is actually coded using maven, you are using ant to compile it.

The downside of using ant is dependency resolution which I think is forte of maven. If maven is used to test the project with travis, you would not need to bundle and ship junit jar file as you are doing currently now. The dependency resolution happens at compile/test time and is different for the 2 phases. So I would recommend against using ant.

Do you like ant for some reason (other than probably legacy)?

243826:
Hi @bartdag, I spent some time today looking at the intermittent test failures. It looks like the race condition within and amongst the various versions of test (total 6 runs - python versions and java versions running together). So the travis CI failures can be written off.

I want to help fix this issue also but would love your buy in and support.

243826:
Also - would really appreciate if you make a release with the changes I submitted so that I can evangelize py4j. Otherwise I am forced to make a local release which many do not appreciate.

bartdag:
Hi, sorry for the delays, I'm currently very busy with work and life in general with the coming holidays :-)

Thanks for your work on Py4J, I'll try to be more responsive.

Regarding this pull request:

1. The tests are indeed tricky and if they are green locally, it's fine with me. I've been trying to improve the test suite's reliability but it is quite challenging.

2. I believe the code still need to be refactored. I would use the name resetCallbackClientAddress, which conveys that there is a reset going on. I do not want to introduce a new terminology (i.e., PythonServer) for now, but I will seriously take this name into consideration when I get rid of the "Gateway/Callback" terminology, hopefully in the near future.

3. Making a release is time consuming, especially deploying to maven central. I'll try to make one next week, but I cannot promise anything.

4. Regarding the use of ant: I went through a few builders (paver, then ant, then someone asked for maven support, etc.). I'm not a huge fan of maven because it seems overly complex for it's purpose and I would much prefer to move to gradle instead of spending time to convert the whole release process from ant to maven. I totally agree that including the junit jar is more a hack and is not necessary with more modern build tools :-)

243826:
@bartdag See my latest commit on this pull request. I think this is the refactoring you wanted to see?

bartdag:
Hi, yes, this looks good. The errors reported by travis are weird though, but it might be time to update to update the travis config for their new build system or to move to circle-ci. As soon as I can try it locally I'll merge the pull request and start to work on a release (there are a few fixes in @master that require a few bits of documentation).

Thanks a lot for your work and your input!

243826:
Thanks!

243826:
@bartdag - I should have tested thoroughly. The resetCallbackClientAddress does not work. It seems that the callback client reference is deeply distributed with the proxies. With the resetCallbackClientAddress, we will need to recall all the references to  CallbackClient being shutdown and it seems complicated. My last commit I added yesterday does not work.

Sorry.

bartdag:
Well my fault for not trying more the feature. Proxies that were started before the reset can fail (similar behaviour to Java objects on the Python side that are accessed after a gateway is shut down). But the thing I did not check is if all sockets are correctly closed during the reset.

243826:
Submitted PR #183.

243826:
BTW, slightly orthogonal, I have a hunch that the java tests fail because they all start on the same machine concurrently and compete for the same ports. On my machine I do serial runs and have yet to see one fail so far (40+).

bartdag:
Travis CI recently changed (again!!) their testing infrastructure and I believe you are right. That's also why I'm looking into circleci. It's slightly less convenient to run these kinds of matrix builds but our experience at work has been more positive than travis.

PR #184, opened 2015-12-29
Fix the issue that PythonProxyHandler.finalize blocks forever

zsxwing:
PythonProxyHandler.finalize uses sendCommand to send the finalizeCommand. Because "CallbackConnection.sendCommand" needs the Python side to send back a return message but the Python side doesn't, it blocks "finalize" forever.

This PR added a "y\n" return message for finalizeCommand to fix this issue.

This issue was found by David Watson in https://issues.apache.org/jira/browse/SPARK-11711

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -1738,8 +1738,10 @@ class CallbackConnection(Thread):
                 elif command == proto.GARBAGE_COLLECT_PROXY_COMMAND_NAME:
                     self.input.readline()
                     del(self.pool[obj_id])
+                    self.socket.sendall("y\n".encode("utf-8"))
                 else:
                     logger.error("Unknown command {0}".format(command))
+                    self.socket.sendall(proto.ERROR_RETURN_MESSAGE.encode("utf-8"))
         except Exception:
             # This is a normal exception...
             logger.info(


bartdag:
Wow! That's a pretty big bug, thanks for the fix. I'll make a small modification to use the existing constant. I will also do another quick pass at all the commands to make sure that this bug is not repeated elsewhere. There is also another open issue related to finalizers (https://github.com/bartdag/py4j/issues/173) and I'll have to raise the priority given the problems that they can cause.

bartdag:
@zsxwing I added a test that fails if the python side does not send a reply. I also added a sotimeout on the Java side so that the finalizer will not block forever waiting for a reply should the python side or the network behave badly. Thanks for the pull request and making me aware of the issue. I was preparing a release for another feature so this fix will be released soonish (if not next week in two weeks).

zsxwing:
Cool. @bartdag I'm wondering if you could also fix #187 and #176 in the next release.

bartdag:
#187 for sure.

For #176 I may have to push it to another release if I want to release 0.9.1 quickly. It's a useful change, but I need to add a test and make sure it correctly mirrors Java exception propagation. I apologize for the long delay.

zsxwing:
No worry about #176. PySpark can catch the exception and pass it to Java side by itself as a workaround. This is already done in PySpark.

PR #189, opened 2016-02-02
Add TLS support

njwhite:
Allow the TCP connections between the py4j python gateway and the JVM to be encrypted (using Python's `ssl` module on the client side, and allowing the server to subclass `GatewayServer` to wrap sockets in whatever `SSLSocketFactory` it chooses to configure.


bartdag:
First, thank you very much for addressing part of the security shortcomings of Py4J (the other being authentication).

If I understand correctly, you would suggest that to use SSL, one would have to subclass GatewayServer and override startSocket() ?

Some documentation, tests, and a corresponding implementation for the callback server (when Java calls Python) would be necessary to move forward with this pull request. I can certainly dedicate some time on this, but if you still have a few cycles to spend on this issue, it would be much appreciated!

njwhite:
@bartdag glad to know I'm on the right lines. I've added callback server support, tests + docs to follow.

njwhite:
@bartdag how about this?

bartdag:
I made a few small comments, but otherwise it looks great, thanks! Do not forget to add your name to the contributor file in the pull request :-)

njwhite:
@bartdag thanks for the detailed review - I've pushed a commit incorporating your suggestions.

bartdag:
Many thanks!

Issue #163, opened 2015-07-15
Add an option to prevent converting Java primitive wrappers to Python int


bartdag:
1. In MethodInvoker.invoke, do not return Object, but InvokerReturnObject which is a tuple containing the return value (Object) and the return value type (Class).

2. In Gateway, accept a GatewayParameter object. For now, we will have only one option, preserveWrapper.

3. When creating a ReturnObject, check preserveWrapper and if it is true, do not convert the wrapper to a primitive.

bartdag:
After this issue is solved, it will be possible to do:
```python
byte_var = gateway.jvm.Byte(2)
short_var = gateway.jvm.Short(3)
isinstance(short_var, JavaObject) # returns True
isintance(short_var, int) # returns False
```

HyukjinKwon:
Is there any workaround for this BTW?

PR #190, opened 2016-02-05
Document Boxing Behaviour

njwhite:
...I was wondering about this


bartdag:
Thanks, this is actually related to #163 and eventually I would like this behavior to be configurable. I'll merge this pull request soon.

PR #191, opened 2016-02-08
Add a "Stream" Type

njwhite:
Add support for sending a stream of bytes (as distinct from a byte[],
which has to be constructed up front) from Java to Python. This
shouldn't require any additional round-trips of the Py4J protocol (e.g.
implementing streaming using an iterator of small byte[] arrays). The
implementation adds another Command to the protocol that allows the
Python application code to access the underlying socket, and read binary
data at its own rate (allowing them to control the marshalling and avoid
any text-based overhead (e.g. the base-64 encoding of byte arrays). The
Java side works based on returning ReadableByteChannel objects from a
Java method, which the Py4J library reads from and writes into its
socket.


py4j-java/src/py4j/commands/Command.java
```diff
@@ -88,9 +88,10 @@ public void execute(String commandName, BufferedReader reader,
 	 * <p>
 	 * Called when a command instance is created and assigned to a connection.
 	 * </p>
-	 * 
+	 *
 	 * @param gateway
+	 * @param connection the {@link GatewayConnection} this socket is assigned to
 	 */
-	public void init(Gateway gateway);
+	public void init(Gateway gateway, GatewayConnection connection);
```
bartdag: can we add an init method instead of changing the signature and breaking backward compatibility?

bartdag:
Hi,

first, thanks for tackling one of the most requested features of Py4J :-)

Because this is such a big feature and it somehow changes the protocol and the API, we'll need to iterate a little bit on this solution before it gets merged:

1. I like the stream() call and the fact that the Python client can then consume the bytes at its own pace. I think it's a better API than the one I was envisioning at first.

2. I can imagine tons of different usage scenarios with little tweaks here and there. We'll probably have to make this extendable at some point, but I'm totally fine with starting with this solution first.

3. As I noted in the pull request, it is important to preserve backward compatibility when possible (the init method).

4. I would greatly prefer to create a new socket to exchange the binary data. The exchange would look like:
    1. python sends a stream command to java
    2. java starts a server socket and responds with the address and port
    3. python receives the response, opens a new client socket and returns a wrapper (connection) 

5. The big advantage is that you could transfer exceptions if they occur instead of just returning an empty response. There are various connection lifecycles that Py4J need to support and stream would break these lifecycles by "locking" the current connection.

6. Alternatively, there could be some exchange at the beginning where the Java side responds with a success or error return message and then starts sending the binary data. I think we can get away with the current connection management code (new connections will be created if necessary) and big documentation warnings (if users creates tons of streams and don't close them, they will create a memory leak), but exceptions occurring on the Java side must be sent to the Python side.

7. Another alternative or addition: a context manager that safely closes the underlying connection:
    ```
    with gateway.stream(...) as stream:
        value = stream.read()
    ```

py4j-java/src/py4j/commands/Command.java
bartdag: please ignore that comment. it was a bit too soon in the morning when I read the code.

You can keep it there, but before making the next release, I'll probably remove the init method altogether and just pass the Gateway and GatewayConnection instances to the execute() method. Since we're breaking backward compatibility, let's break it for a good cause :-)

Unless I'm mistaken, if we send the GatewayCOnnection instance to the execute method, all commands would not need to keep any state at all, which would definitively improve the design.

njwhite:
@bartdag thanks for the response. I agree there's a lot of iteration to go, but I've pushed a changing incorporating some of your suggestions. In particular:
4, 5, 6) I definitely agree exceptions are important. However, I much prefer solution (6) as I'm running this in a very locked down environment, so my Python and Java processes can't just start listening on arbitary ports or make outbound connections due to restrictive firewall rules. I've therefore modified the PR to send a success or error message in the Py4J protocol as the first (text) line of the response, and use `get_return_value` to unwrap (and possibly throw) it as usually. The binary data from the returned `ReadableByteChannel` is dumped into the socket as before after the Protocol.END_OUTPUT is flushed.
7) Error guards are a good idea - the returned GatewayConnection is now wrapped in a guard that returns it to the connection pool once finished.

bartdag:
Hi, thanks for your changes, that's greatly appreciated! I will review them in the coming days.

njwhite:
Hi @bartdag - I was wondering if you'd had a chance to review this approach yet. Thanks -

bartdag:
Hi, apologies for the delay. I aim to test the branch locally and merge it this week if I don't see any issues.

bartdag:
Hi, thanks a lot for this major contribution. I was not familiar with the Channel API (especially the rewind/flip/compact protocol), and it would have taken me many more attempts to get it right.

In your experience, is there any performance benefits in increasing the buffer (e.g., on the Java side) for large data transfer?

This new feature will be part of the 0.10 release, which will also include my work on a new threading model.

PR #196, opened 2016-03-11
Use current thread's context classloader in Class.forName() calls

JoshRosen:
This patch modifies Py4J's Java code to replace all `Class.forName` calls with calls to a static helper method which uses the current thread's context class loader to load classes. 

`Class.forName(className)` loads classes from the root class loader and thus will miss classes that are only available in child class loaders. This has caused problems for us in Apache Spark; see https://issues.apache.org/jira/browse/SPARK-6047 for more details.

This is related to #130 but addresses a narrower set of issues.

### DIFF
--- a/py4j-java/src/py4j/Protocol.java
+++ b/py4j-java/src/py4j/Protocol.java
@@ -36,6 +36,7 @@ import java.lang.reflect.Proxy;
 import java.math.BigDecimal;
 
 import py4j.reflection.PythonProxyHandler;
+import py4j.reflection.ReflectionUtils;
 
 /**
  * <p>
@@ -401,7 +402,7 @@ public class Protocol {
 
 		for (int i = 1; i < length; i++) {
 			try {
-				interfaces[i - 1] = Class.forName(parts[i]);
+				interfaces[i - 1] = ReflectionUtils.classForName(parts[i]);
 				if (!interfaces[i - 1].isInterface()) {
 					throw new Py4JException(
 							"This class "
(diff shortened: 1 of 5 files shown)


JoshRosen:
@bartdag, if we merge this fix in then I'd like to be able to upgrade Spark to use a new published version of Py4J which incorporates this patch.

While I suspect that we'll want to upgrade to the next Py4J release, I'm a bit concerned about the fact that it contains so many changes. As a result, would it be possible to publish (to Maven) a 0.9.2 or 0.9.1.1 release which includes only this additional patch?

bartdag:
@JoshRosen thanks for the patch: I'll eventually add a way to register custom class loading strategies to suit more complex needs (e.g., OSGi), and move away from the static method so different gateways can have different class loading strategies.

Regarding the release: I'll create a branch from 0.9.1 and add this change to 0.9.2, which I'll publish to maven/pypi. I'll change the code in master to use 0.10-SNAPSHOT.

Regarding the future: I would like to run by you or your colleagues the list of breaking changes I intend to make before Py4J 1.0. I am not making these changes lightly, but I believe this will make it a lot easier to maintain Py4J and support the increasing variety of usage scenarios. It would be very helpful to learn which changes could be problematic for you and if you have an idea of a timeline to integrate such changes. Can I send you an email about that?

Thanks!

JoshRosen:
@bartdag, that release strategy sounds like a good plan; thanks for helping out with this.

Let's definitely chat more Py4J 1.0; my email is joshrosen@databricks.com. Given a rough outline of the planned changes, I can help to audit our codebases to figure out how heavily we'll be impacted.

bartdag:
The branch to follow for the 0.9.x release is [stable/0.9.x](https://github.com/bartdag/py4j/tree/stable/0.9.x)

bartdag:
Py4J 0.9.2 has been released on pypi et maven central.

JoshRosen:
Thanks a bunch for helping with this. I've opened https://github.com/apache/spark/pull/11687 to fix the longstanding PySpark bug that motivated this patch.

PR #204, opened 2016-05-17
Add a Builder object for ClientServer

jonahkichwacoders:
As ClientServer's constructor has very many arguments, using a Builder
for constructing one makes code easier to read.

### DIFF
--- a/py4j-java/src/main/java/py4j/ClientServer.java
+++ b/py4j-java/src/main/java/py4j/ClientServer.java
@@ -160,4 +160,85 @@ public class ClientServer {
 		return proxy;
 	}
 
+	/**
+	 * Helper class to make it easier and self-documentating how a
+	 * {@link ClientServer} is constructed.
+	 */
+	public static class ClientServerBuilder {
+		private int javaPort;
+		private InetAddress javaAddress;
+		private int pythonPort;
+		private InetAddress pythonAddress;
+		private int connectTimeout;
+		private int readTimeout;
+		private ServerSocketFactory serverSocketFactory;
+		private SocketFactory socketFactory;
+		private Object entryPoint;
+
+		public ClientServerBuilder() {
+			this(null);
+		}
+
+		public ClientServerBuilder(Object entryPoint) {
+			javaPort = GatewayServer.DEFAULT_PORT;
+			javaAddress = GatewayServer.defaultAddress();
+			pythonPort = GatewayServer.DEFAULT_PYTHON_PORT;
+			pythonAddress = GatewayServer.defaultAddress();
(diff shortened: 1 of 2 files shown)


py4j-java/src/main/java/py4j/ClientServer.java
```diff
@@ -160,4 +160,85 @@ public Object getPythonServerEntryPoint(Class[] interfacesToImplement) {
 		return proxy;
 	}
 
+	/**
+	 * Helper class to make it easier and self-documentating how a
+	 * {@link ClientServer} is constructed.
+	 */
+	public static class ClientServerBuilder {
+		private int fJavaPort;
```
bartdag: Could you please drop the f in front of the field name (this would be the only place in Py4J where we do that) :-)

py4j-java/src/main/java/py4j/ClientServer.java
```diff
@@ -160,4 +160,85 @@ public Object getPythonServerEntryPoint(Class[] interfacesToImplement) {
...
+			fPythonPort = GatewayServer.DEFAULT_PYTHON_PORT;
+			fPythonAddress = GatewayServer.defaultAddress();
+			fConnectTimeout = GatewayServer.DEFAULT_CONNECT_TIMEOUT;
+			fReadTimeout = GatewayServer.DEFAULT_READ_TIMEOUT;
+			fServerSocketFactory = ServerSocketFactory.getDefault();
+			fSocketFactory = SocketFactory.getDefault();
+			fEntryPoint = entryPoint;
+		}
+
+		public ClientServer build() {
```
bartdag: Is it possible to add a small unit test demonstrating (and testing) the use of the builder?

jonahkichwacoders:
Yes, I will do those things. I have moved that class from EASE (where I wrote it) to Py4J, but of course EASE has the opposite requirement, everything must be f'ed.

jonahkichwacoders:
I have added a test, but as most of the parameters are hard to test I only tested one based on py4j.GatewayServerTest.testEphemeralPort()

Note that I need to do a downcast from Py4JJavaServer to GatewayServer to get the port. Perhaps getListeningPort should be added to Py4JJavaServer?

bartdag:
Thanks for the test. I'll see if there are other methods missing in Py4JJavaServer and add them. The downcast should be unnecessary: good catch!

Regarding the pull request, if you hava Java 8, you should run ``./gradlew spotlessApply`` to format your code (the build error is about formatting rule violations). If you don't have Java 8 nearby, I'll run it before merging.

jonahkichwacoders:
OK, I will do that Java 8 bit soon.

bartdag:
Thanks!

PR #205, opened 2016-05-20
Make ClientServerConnections per thread-per client-server instance

jonahkichwacoders:
This allows a single JVM or Python to have multiple ClientServer objects
instantiated and used from the same thread.

I decided to use a Map/dict of integer (incremented per PythonClient/JavaClient) to connection objects. 
Three reasons of differing validity:
- avoided changing thread local to non-static scope
- no additional references to PythonClient/JavaClient in "hard to see" thread locals
- PythonClient/JavaClient were not obviously guaranteed to stay hashable

There is a downcast in ClientServerConnection.setThreadConnection that could use a review.

### DIFF
--- a/py4j-java/src/main/java/py4j/ClientServerConnection.java
+++ b/py4j-java/src/main/java/py4j/ClientServerConnection.java
@@ -44,7 +44,11 @@ public class ClientServerConnection implements Py4JServerConnection, Py4JClientC
 
 	private boolean used = false;
 	private boolean initiatedFromClient = false;
-	private static ThreadLocal<ClientServerConnection> threadConnections = new ThreadLocal<ClientServerConnection>();
+	private static ThreadLocal<Map<Integer, ClientServerConnection>> threadConnections = new ThreadLocal<Map<Integer, ClientServerConnection>>() {
+		protected java.util.Map<Integer, ClientServerConnection> initialValue() {
+			return new HashMap<Integer, ClientServerConnection>();
+		};
+	};
 	protected Socket socket;
 	protected BufferedWriter writer;
 	protected BufferedReader reader;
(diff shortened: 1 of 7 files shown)


jonahkichwacoders:
Still learning on the build stuff. I'm fixing some flake8 failures and others and will resubmit soonish

bartdag:
Hi Jonah, I'll take a closer look tonight, but unless I missed something, one issue with a static threadlocal is that memory is not reclaimed when a server is shut down (the connection stays in the map). We can probably try to remove the connection when the server shuts down, but that's also why I wanted to investigate an instance-level threadlocal (less common, but automatic memory management).

bartdag:
Thanks for the elaborate test case and a potential solution, that's greatly appreciated.

jonahkichwacoders:
Feel free to use or dispose whichever parts are useful to you. AFAICT there is leak, in this version and the master. I'm not sure how to write the test case to verify that one though.

jonahkichwacoders:
Ran into some Python3 vs Python2 because Python3 does not have a long type anymore. So I bottled it and used strings to pass thread ids from Python to Java. Now all checks pass, so I guess it is ready for review.

jonahkichwacoders:
The places that are not covered in the Java code are covered by the Python tests I added. I basically found it too complicated to try and write those new tests in just Java.

bartdag:
Regarding the coverage: the java side is indeed missing testing utilities and most Java code is actually tested from Python. Totally fine.

Regarding the memory leak, the master branch contains a leak, but it's less severe. A thread can only refer to one connection and when the thread dies, the reference is removed. With a map, a long-lived thread will accumulate references without ever releasing them. Manually releasing the references can also be tricky because a server might be shut down by another thread. I'll reuse the test, but I'll see if I can implement the threadlocal variable per instance.

bartdag:
I tried the per-instance approach [on branch threadlocal-per-instance](https://github.com/bartdag/py4j/tree/threadlocal-per-instance). All tests (including your new test) passes. I'll have to work on the interface to prevent downcasting and I'll perform a few memory tests to make sure that the connections are correctly garbage collected (I believe the swt ui thread might still keep a reference to the threadlocal even if the PythonClient is garbage collected... that would be bad). I'll also add a test where two python threads communicate with two java threads through the same pair of port.

I really like your multi client server test btw :-)

bartdag:
Just a quick status update: I conducted several experiments and found that the current usage of threadlocals (either static or per-instance) makes it extremely difficult and error-prone to clean up the various connection and server instances when multiple clientservers are created by long-lived threads. There are workarounds, but nothing robust enough to support weird but potential use cases. For example, if you have multiple long-lived threads (say, a worker pool + the swt ui thread) communicating with the same clientservers, each thread would have to call clientserver.shutdown() to make sure that the threads release their references to the threadlocals/clientserverconnections.

I considered two solutions: (1) swapping threadlocals with a concurrent hash map indexed by threads (it seems this was how threadlocals were implemented or simulated in the early days) or (2) using weak references in the existing thread locals.

(1) has a performance penalty, but (2) seems very promising from my tests, although it requires a few small refactorings.

bartdag:
Hi Jonah, I merged into master the fix for this and reused your unit test. Let me know if everything works for you and if we can close this pull request.

Thanks again for your help!

jonahkichwacoders:
OK, I'll close this out and raise new issues/pulls if needed. As it passes my multi thread test I am pretty confident.

PR #228, opened 2016-06-14
Detect address type (IPv4/IPv6) and connect with good AF

kujiu:
Hello,

This fix make Py4J compatible with IPv6. Can you merge it on your branch or says me what to fix for your requirements please ?

Thanks for all in advance.

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -795,7 +795,8 @@ class GatewayConnection(object):
         self.gateway_parameters = gateway_parameters
         self.address = gateway_parameters.address
         self.port = gateway_parameters.port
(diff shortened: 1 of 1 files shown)


bartdag:
Thanks a lot, I'll take a closer look next week. Maybe two small points in the meantime:

1. I guess the Java side already supports IP v6 because it uses InetAddress?
2. Could you add a test demonstrating that ip v6 works. This will somehow require a modification to the Java side, but that would really help.

bartdag:
Quick update: as you can see from my commits, this is unfortunately not as simple as expected. For instance, Python 2.6 does not support IPv6 and the CI environment we use (CircleCI) does not enable by default IPv6. I'll think I caught most corner cases, but I'll have to continue tomorrow.

bartdag:
Thanks for initiating this patch. I added you to the AUTHORS.txt list.

PR #233, opened 2016-06-30
Use DynamicImport-Package in MANIFEST.MF

jonahkichwacoders:
As suggested by Scott Lewis:

Using DynamicImport-Package instead of Eclipse-BuddyPolicy means that
the directive applies to all OSGi frameworks, instead of just
Eclipse/Equinox ones.


bartdag:
Thanks! Merged

PR #250, opened 2016-08-09
Fixed NPE that occurs when creating a SWT gateway connection

PeterC-DLS:
There is a possibility that custom commands has not been initialized


jonahkichwacoders:
@PeterC-DLS Can you track down how customCommands became null? AFAICT it should not be possible, or at least isn't the intention reading the constructor which has a null check already when initializing customCommands.

PeterC-DLS:
It was a fair while back so the code may have changed to prevent it! I'm just porting some old commits so I guess this one does not apply any longer

PR #251, opened 2016-08-09
Support enumerations

PeterC-DLS:
Support methods that have enumeration arguments


bartdag:
Hi, thanks for your pull request. I'm not sure what the purpose is though. Enum types have been supported since 0.8, but without the string to enum conversion.

For example, if you have something like that:

```java
import com.package.PlanetEnumType;
...
public String myMethod(PlanetEnumType planet) {
    return enumArg.toString();
}
...
```

You can do this:
```python
mercury = gateway.jvm.com.package.PlanetEnumType.MERCURY

my_java_object.myMethod(mercury)
```

PeterC-DLS:
Hi Bart, I didn't realise as I'm porting some of our changes. Will close this, thanks.

PR #267, opened 2017-01-10
Add path to fix #266

### LINKED ISSUE #266: Could not find py4j jar when installed with pip install --user
When py4j is installed using `pip install --user py4j` (pip version 8.1.1, python 2.7.12, as installed on Ubuntu 16.04), I get the following error:
```
>>> import py4j.java_gateway as jg
>>> port = jg.launch_gateway()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dcs/.local/lib/python2.7/site-packages/py4j/java_gateway.py", line 270, in launch_gateway
    raise Py4JError("Could not find py4j jar at {0}".format(jarpath))
py4j.protocol.Py4JError: Could not find py4j jar at 
```

davidcsterratt:
Adding this path allows the jar to be found.

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -194,6 +194,9 @@ def find_jar_path():
         os.path.realpath(__file__)), "../share/py4j/" + jar_file))
     paths.append("../../../current-release/" + jar_file)
     paths.append(os.path.join(sys.prefix, "share/py4j/" + jar_file))
+    # pip install --user
+    paths.append(os.path.join(os.path.dirname(
+            os.path.realpath(__file__)), "../../../../share/py4j/" + jar_file))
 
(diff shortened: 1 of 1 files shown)


bartdag:
Thanks for the bug report and the contribution, one thing that would help future maintenance and I've been bad at keeping up with this, is to add in a comment the expected directory structure, something like:

```python
# pip --user
# this file is here: virtualenvpath/lib/pythonX/site-packages/py4j/java_gateway.py
# the jar file is here: virtualenvpath/share/py4j/py4j.jar
```

Thanks!

davidcsterratt:
Good idea - comments added as requested. I've tested that the extra path works with installs in /usr/local (default location) and ~/.local (with --user option).

bartdag:
merged

PR #279, opened 2017-06-21
Add failing test for deep inherited interfaces

gatesn:


bartdag:
Thanks, merged when #280 was fixed.

Issue #275, opened 2017-05-31
Client/Server transport corruption caused by py4j memory management 


jonahkichwacoders:
Hi Barthelemy,

I have been having occasional random exceptions with py4j (and its integration in Eclipse EASE) like the following:
```
py4j.Py4JException: An exception was raised by the Python Proxy. Return Message: x
	at py4j.Protocol.getReturnValue(Protocol.java:438)
	at py4j.reflection.PythonProxyHandler.invoke(PythonProxyHandler.java:105)
	at com.sun.proxy.$Proxy25.executeScript(Unknown Source)
	at org.eclipse.ease.lang.python.py4j.internal.Py4jScriptEngine.internalExecute(Py4jScriptEngine.java:221)
	at org.eclipse.ease.lang.python.py4j.internal.Py4jScriptEngine.execute(Py4jScriptEngine.java:211)
	at org.eclipse.ease.AbstractScriptEngine.inject(AbstractScriptEngine.java:228)
	at org.eclipse.ease.AbstractScriptEngine.run(AbstractScriptEngine.java:289)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:55)
```

As it turns out this is caused by py4j memory management sending a memory delete (MEMORY_DEL_SUBCOMMAND_NAME) to the Java in the middle of receiving a command from Java. 

Here is an extract of a log, with comments (#)

```
May 31, 2017 12:38:57 PM py4j.ClientServerConnection sendCommand
INFO: Sending Python command: c  # Java starts calling Python
t
sendObject
ro96
ro97
e

Received command c on object id t  # Python receiving the call
Command to send: m  # in the middle of processing the call from Java, Python garbage collection start
d
o0
e

Answer received: e  # this e is the end of the Java call (after ro97)
Unknown command e
May 31, 2017 12:38:57 PM py4j.ClientServerConnection sendCommand
INFO: Returning CB command: x
py4j.Py4JException: An exception was raised by the Python Proxy. Return Message: x
	at py4j.Protocol.getReturnValue(Protocol.java:449)
	at py4j.reflection.PythonProxyHandler.invoke(PythonProxyHandler.java:109)
	at com.sun.proxy.$Proxy0.sendObject(Unknown Source)
	at py4j.examples.SingleThreadClientApplicationSendsObjects.main(SingleThreadClientApplicationSendsObjects.java:48)
```

[...]

Issue #243, opened 2016-07-21
Make ClassLoadingStrategy specific to GatewayServer instances


bartdag:
Currently, it is only possible to set a global ClassLoadingStrategy, but it would be more flexible to set a strategy per GatewayServer instance.

scottslewis:
I've been playing around with the code and I have some thoughts I want to propose and discuss.

First, we could add a ClassLoadingStrategy param to the GatewayServer constructor...e.g.

	public GatewayServer(Object entryPoint, int port, InetAddress address, int connectTimeout, int readTimeout,
			List<Class<? extends Command>> customCommands, Py4JPythonClient cbClient,
			ServerSocketFactory sSocketFactory, ClassLoadingStrategy classLoadingStrategy) {

Then the Gateway instance (created in the GatewayServer constructor could be passed this classLoadingStrategy...e.g.

		this.gateway = new Gateway(entryPoint, cbClient, classLoadingStrategy);

and the classLoadingStrategy would be stored as a field in the Gateway instance.  Also, create a method on Gateway.getClassLoadingStrategy() that returns the member.

Now, in the Protocol.getPythonProxy method could be enhanced to optionally use this ClassLoadingStrategy, falling back to the ReflectionUtil classLoadingStrategy if there is none set for the Gateway:

	public static Object getPythonProxy(String commandPart, Gateway gateway) {
		String proxyString = commandPart.substring(1, commandPart.length());
		String[] parts = proxyString.split(";");
		int length = parts.length;
		Class<?>[] interfaces = new Class<?>[length - 1];
		if (length < 2) {
			throw new Py4JException("Invalid Python Proxy.");
		}

		ClassLoadingStrategy classLoadingStrategy = gateway.getClassLoadingStrategy();
		if (classLoadingStrategy == null) 
			classLoadingStrategy = ReflectionUtil.getClassLoadingStrategy();
		// Just in case
		if (classLoadingStrategy == null)
			classLoadingStrategy = new CurrentThreadClassLoadingStrategy();
		
		for (int i = 1; i < length; i++) {
			try {
				interfaces[i - 1] = classLoadingStrategy.classForName(parts[i]);
				if (!interfaces[i - 1].isInterface()) {
					throw new Py4JException(
							"This class " + parts[i] + " is not an interface and cannot be used as a Python Proxy.");
				}
			} catch (ClassNotFoundException e) {
				throw new Py4JException("Invalid interface name: " + parts[i]);
			}
		}

[...]

PR #281, opened 2017-06-23
Fix for issue https://github.com/bartdag/py4j/issues/278

scottslewis:
PythonProxyHandler class (java).  Also enhanced java_callback_test.py to
test.  Without fix to PythonProxyHandler new test fails, with fix
succeeds.  see IntegrationTest.testProxyReturnerNull

### DIFF
--- a/py4j-python/src/py4j/tests/java_callback_test.py
+++ b/py4j-python/src/py4j/tests/java_callback_test.py
@@ -119,6 +119,9 @@ class Returner(object):
     def doNothing(self):
         print("Doing nothing")
 
+    def getNull(self):
+        return None
+    
     class Java:
         implements = ["py4j.examples.IReturnConverter"]
 
@@ -317,6 +320,13 @@ class IntegrationTest(unittest.TestCase):
         returner = Returner()
         output = example.computeNothing(returner)
         self.assertEqual(output, 1)
+
+    def testProxyReturnerNull(self):
+        sleep()
+        example = self.gateway.jvm.py4j.examples.ReturnerExample()
+        returner = Returner()
+        output = example.computeNull(returner)
+        self.assertEqual(output, None)
 
     def testProxy(self):
         sleep()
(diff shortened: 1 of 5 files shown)


py4j-python/src/py4j/tests/java_callback_test.py
```diff
@@ -318,6 +321,13 @@ def testProxyReturnerVoid(self):
         output = example.computeNothing(returner)
         self.assertEqual(output, 1)
 
+    def testProxyReturnerNull(self):
+        sleep()
+        example = self.gateway.jvm.py4j.examples.ReturnerExample()
+        returner = Returner()
+        output = example.computeNull(returner)
+        self.assertEqual(output, None)
```
bartdag: I was about to suggest to use assertIsNone, but it doesn't exist in Python 2.6, so assertEqual it is!

bartdag:
Thanks for the pull request, it is almost good to go :-) Can you add your name to AUTHORS.txt?

scottslewis:
Hi @bartdag .   OK added myself to authors.txt with second commit for this pull request.   Thanks for merging this quickly, it will be nice to have in 0.10.6.   

What's your thinking on timing for 0.10.6?   

Would it be possible to deal with:  https://github.com/bartdag/py4j/issues/243  ?   I could help with impl 
 and testing if you like.   The reason this is important is that for my use case there may be cases when there may be more than one gateway, and each gateway should have it's own classloading strategy (since there is more than one classloader in OSGi).

bartdag:
Regarding #243, do not hesitate to open a pull request. I can take the work from there (e.g., port the fix to ClientServer). I don't think I'll have time to work on it in the coming weeks otherwise.

My focus for 0.10.6 is #275 

I should be able to work on it next Friday. If I can complete it, I'll be able to release 0.10.6 during the next weekend.

Thanks for your pull request! I've squashed and merge it.

PR #282, opened 2017-06-27
Add GatewayServer constructor to allow for customized Gateway (osgi)

scottslewis:
For usage in OSGi, it will be very helpful to add a GatewayServer constructor with the following signature:

	public GatewayServer(Gateway gateway, int port, InetAddress address, int connectTimeout, int readTimeout,List<Class<? extends Command>> customCommands, ServerSocketFactory sSocketFactory) 

This will allow the use of an subclass of the Gateway class, so that the Gateway may be customized.  One way that this will be used in OSGi environment is to subclass Gateway to allow for the customization of proxy creation...doing the 'appropriate thing'...wrt classloading and etc...from within OSGi environment (where there is more than one relevant classloader).

This is a new GatewayServer constructor/entry point so is very unlikely to introduce problems with existing code.  All tests pass with this addition.  There are a couple small additions of annotations to prevent compiler warnings (@SuppressWarnings)

### DIFF
--- a/py4j-java/src/main/java/py4j/Gateway.java
+++ b/py4j-java/src/main/java/py4j/Gateway.java
@@ -30,6 +30,7 @@
 package py4j;
 
 import java.lang.reflect.Array;
+import java.lang.reflect.Proxy;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.InetAddress;
@@ -43,6 +44,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Logger;
 
 import py4j.reflection.MethodInvoker;
+import py4j.reflection.PythonProxyHandler;
 import py4j.reflection.ReflectionEngine;
 
 /**
@@ -349,6 +351,13 @@ public class Gateway {
 		this.isStarted = isStarted;
 	}
 
+	public Object createProxy(ClassLoader classLoader, @SuppressWarnings("rawtypes") Class[] interfacesToImplement, String objectId) {
+		return Proxy.newProxyInstance(classLoader, interfacesToImplement, createPythonProxyHandler(objectId));
+	}
+	
+	protected PythonProxyHandler createPythonProxyHandler(String id) {
+		return new PythonProxyHandler(id, this);
+	}
 	/**
 	 * <p>
 	 * Releases all objects that were referenced by this Gateway and shuts
--- a/py4j-java/src/main/java/py4j/Gateway.java
+++ b/py4j-java/src/main/java/py4j/Gateway.java
@@ -43,7 +43,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Logger;
 
-import py4j.reflection.ClassLoadingStrategy;
 import py4j.reflection.MethodInvoker;
 import py4j.reflection.PythonProxyHandler;
 import py4j.reflection.ReflectionEngine;
@@ -392,10 +391,6 @@ public class Gateway {
 			bindings.put(Protocol.ENTRY_POINT_OBJECT_ID, entryPoint);
 		}
 		bindings.put(Protocol.DEFAULT_JVM_OBJECT_ID, defaultJVMView);
-	}
-
-	public ClassLoadingStrategy getClassLoadingStrategy() {
-		return null;
 	}
 
 }
--- a/py4j-java/src/main/java/py4j/Gateway.java
+++ b/py4j-java/src/main/java/py4j/Gateway.java
@@ -43,6 +43,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Logger;
 
+import py4j.reflection.ClassLoadingStrategy;
 import py4j.reflection.MethodInvoker;
 import py4j.reflection.PythonProxyHandler;
 import py4j.reflection.ReflectionEngine;
@@ -391,6 +392,10 @@ public class Gateway {
 			bindings.put(Protocol.ENTRY_POINT_OBJECT_ID, entryPoint);
 		}
 		bindings.put(Protocol.DEFAULT_JVM_OBJECT_ID, defaultJVMView);
+	}
+
+	public ClassLoadingStrategy getClassLoadingStrategy() {
+		return null;
 	}
 
 }
--- a/py4j-java/src/main/java/py4j/GatewayServer.java
+++ b/py4j-java/src/main/java/py4j/GatewayServer.java
@@ -427,6 +427,35 @@ public class GatewayServer extends DefaultGatewayServerListener implements Py4JJ
 		this.sSocketFactory = sSocketFactory;
 	}
 
+	/*
+	 * @param gateway gateway instance (or subclass).  Must not be <code>null</code>.
+	 * @param port the host port to usu
+	 * @param address the host address to use
+	 * @param connectTimeout the connect timeout (ms)
+	 * @param readTimeout the read timeout (ms)
+	 * @param customCommands any customCommands to use.  May be <code>null</code>
+	 * @param sSocketFactory socketFactory to use.  Must not be <code>null</code>
+	 */
+	public GatewayServer(Gateway gateway, int port, InetAddress address, int connectTimeout, int readTimeout,
+			List<Class<? extends Command>> customCommands, ServerSocketFactory sSocketFactory) {
+		super();
+		this.port = port;
+		this.address = address;
+		this.connectTimeout = connectTimeout;
+		this.readTimeout = readTimeout;
+		this.gateway = gateway;
+		this.pythonPort = gateway.getCallbackClient().getPort();
+		this.pythonAddress = gateway.getCallbackClient().getAddress();
+		this.gateway.putObject(GATEWAY_SERVER_ID, this);
+		if (customCommands != null) {
+			this.customCommands = customCommands;
+		} else {
+			this.customCommands = new ArrayList<Class<? extends Command>>();
+		}
+		this.listeners = new CopyOnWriteArrayList<GatewayServerListener>();
+		this.sSocketFactory = sSocketFactory;
+	}
+
 	public void addListener(GatewayServerListener listener) {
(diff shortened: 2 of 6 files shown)


py4j-java/src/main/java/py4j/GatewayServer.java
```diff
@@ -754,7 +783,7 @@ protected void startSocket() throws Py4JNetworkException {
 	 * @param interfacesToImplement
 	 * @return
 	 */
-	public Object getPythonServerEntryPoint(Class[] interfacesToImplement) {
+	public Object getPythonServerEntryPoint(@SuppressWarnings("rawtypes") Class[] interfacesToImplement) {
```
bartdag: does the method parameter annotation works for Java 6 and 7? I thought it was a Java 8 thing?

py4j-java/src/main/java/py4j/GatewayServer.java
```diff
@@ -427,6 +427,35 @@ public GatewayServer(Object entryPoint, int port, InetAddress address, int conne
...
+	/*
+	 * @param gateway gateway instance (or subclass).  Must not be <code>null</code>.
+	 * @param port the host port to usu
+	 * @param address the host address to use
+	 * @param connectTimeout the connect timeout (ms)
+	 * @param readTimeout the read timeout (ms)
+	 * @param customCommands any customCommands to use.  May be <code>null</code>
+	 * @param sSocketFactory socketFactory to use.  Must not be <code>null</code>
+	 */
+	public GatewayServer(Gateway gateway, int port, InetAddress address, int connectTimeout, int readTimeout,
```
bartdag: it might be best to create a constructor that is a superset of ```	public GatewayServer(Object entryPoint, int port, InetAddress address, int connectTimeout, int readTimeout,
			List<Class<? extends Command>> customCommands, Py4JPythonClient cbClient,
			ServerSocketFactory sSocketFactory) {```

And add the `Gateway` parameter to the `GatewayServerBuilder`

bartdag:
This looks good (I only mentioned two points in my code review). Just to make sure, this does not completely address #243 right?

py4j-java/src/main/java/py4j/GatewayServer.java
scottslewis: Hi Barthelemy,

It says since 1.5 for the @SuppressWarnings annotation here:

https://docs.oracle.com/javase/7/docs/api/java/lang/SuppressWarnings.html

Although I admittedly haven't tested against either 1.6 or 1.7.

py4j-java/src/main/java/py4j/GatewayServer.java
scottslewis: Hi Barthelemy,

I would be OK with this, although I think it could be confusing having a Gateway parameter to GatewayServerBuilder...since Gateway incorporates a number of others.  But your call.  As long as it's possible to define a subclass of Gateway and have it used for GatewayServer.   It's a little unfortunate that GatewayServer already has many params, and since it's API you can't really change that at this point without breaking many clients/API.

scottslewis:
Hi Barthelemy.   No it does not address #243.   It does allow workarounds, but I'm going to try to fully address #243 tomorrow/wed with another pull request

py4j-java/src/main/java/py4j/GatewayServer.java
bartdag: I was wondering about the method parameter annotation and not SupressWarnings in particular, but it seems that it is indeed supported (CircleCI will catch it anyway).

py4j-java/src/main/java/py4j/GatewayServer.java
bartdag: It is indeed unfortunate and it clearly shows the evolution of the API throughout the years :-) I think I would still prefer to include any new parameters in GatewayServerBuilder. As you say, Gateway already incorporates some of the constructor parameters so if one were to supply a Gateway instance, an entryPoint and a Py4JPythonClient instance, the Gateway instance would take priority and the other two parameters would be ignored (Javadoc could clarify this unfortunate case).

Maybe we can keep two constructors: the one you defined which is the sensible one, and the other that will be used by the builder. At this point, the builder makes more sense for many users considering the large number of parameters (and use cases!).

bartdag: On second thought, you could keep constructor you created and put the decision logic in the builder (if a gateway field is not null, use the constructor accepting a gateway and ignore the callback client and the entry point fields).

scottslewis:
Barthelemy I just attached to this pull request the fix for #243.   The commit description describes what it does.   Now this pull request can be used to resolve both this and #243.  Basically it delegates the class loading strategy to the Gateway.getClassLoadingStrategy() (new method), and the default impl (in Gateway class) is to return null.   In that case (no override) it falls back to use the ReflectionUtil classloading strategy, i.e. the one classloading strategy for all gateways.

py4j-java/src/main/java/py4j/Protocol.java
```diff
@@ -402,9 +402,16 @@ public static Object getPythonProxy(String commandPart, Gateway gateway) {
 			throw new Py4JException("Invalid Python Proxy.");
 		}
 
+	    ClassLoadingStrategy classLoadingStrategy = gateway.getClassLoadingStrategy();
+	    if (classLoadingStrategy == null) 
+	        classLoadingStrategy = ReflectionUtil.getClassLoadingStrategy();
+	    // Just in case
+	    if (classLoadingStrategy == null)
```
bartdag: I would remove this: by default there is a class loading strategy. If we want to prevent against a user error (setting the class loading strategy to Null, we could validate the parameter in `ReflectionUtil.setClassLoadingStrategy` instead.

py4j-java/src/main/java/py4j/Gateway.java
```diff
@@ -349,6 +352,13 @@ public void setStarted(boolean isStarted) {
 		this.isStarted = isStarted;
 	}
 
+	public Object createProxy(ClassLoader classLoader, @SuppressWarnings("rawtypes") Class[] interfacesToImplement, String objectId) {
```
bartdag: If you could add javadoc for the two new public methods, that would be great!

py4j-java/src/main/java/py4j/Protocol.java
bartdag: Also, looking at the code, `ReflectionUtil.classForName` and `ReflectionUtil.getClassLoader` is used in more places this this one (see `ReflectionEngine.getConstructor/getField/getMethod`, `TypeUtil.forName/getClass/isInstanceOf`, `HelpPageCommand.getHelpClass`). 

There are many strategies (e.g., overload `ReflectionUtil.classForName` with an additional classForName method that takes a Gateway instance as a second parameter), but methods in `ReflectionUtil` and `TypeUtil` will be the most difficult to refactor because they don't currently have access to the gateway.

py4j-java/src/main/java/py4j/Protocol.java
```diff
@@ -414,14 +421,7 @@ public static Object getPythonProxy(String commandPart, Gateway gateway) {
...
 
-		Object proxy = getPythonProxyHandler(ReflectionUtil.getClassLoader(), interfaces, parts[0], gateway);
-
-		return proxy;
-	}
-
-	public static Object getPythonProxyHandler(ClassLoader classLoader, Class[] interfacesToImplement, String objectId,
-			Gateway gateway) {
-		return Proxy.newProxyInstance(classLoader, interfacesToImplement, new PythonProxyHandler(objectId, gateway));
+		return gateway.createProxy(ReflectionUtil.getClassLoader(), interfaces, parts[0]);
```
bartdag: shouldn't this use the classLoadingStrategy variable declared above?

bartdag:
Hi Scott, I reviewed the code and identified many more places where a non-trivial change is required (TypeUtil and ReflectionUtil). This is clearly not an easy fix :-( Thanks for your work on this issue!

scottslewis:
Hi barthelemy.  I'm going to have limited time to work on this more this
week so perhaps we should fall back for 10.6.  I would like to get the
gatewayserver constructor work for gateway extension in though if at all
possible for 10.6.  I'll take a look at where else ReflectionUtil is used
today if at all possible.

scottslewis:
Hi Barthelemy. As you can see I've reverted the commit that has the change/fix for but #243.   As you say, with TypeUtil and ReflectionUtil static methods, it's going to be more involved to delegate these functions to an individual gateway instance.   With the ability to extend Gateway, however, it will be possible for a coordinator (responsible for creating Gateways) to define the classloading strategy...that includes the appropriate classloader/namespace scoping for the entire process in (e.g.) OSGi environments.   This will serve for my OSGi remote services impl because it creates it's own proxy for the remote service...following the classloading/versioning/etc rules as per OSGi spec...and then uses the Py4j proxy underneath.   And as I said, the thing that creates multiple gateways can set the necessary py4j classloading strategy appropriately using the BundleClassloadingStrategy I've created.

So...let's forget about dealing with #243 for now, and just get the ability to extend the Gateway into place.    If this much could be done for 0.10.6 it would be extremely helpful for me, ECF remote services, and for users of RS.

If you want to add on a new package (e.g. org.py4j.osgi) in future versions of py4j, I'll be happy to contribute some of these classes:

https://github.com/ECF/Py4j-OSGi/tree/master/bundles/org.py4j.osgi

bartdag:
Hi Scott, thanks for the work and the revert. I'm still waiting for Jonah to test a fix. He said he would do it next week. What kind of deadline for 0.10.6 are you looking for?

scottslewis:
WRT 0.10.6...next week is fine.   Thanks.

bartdag:
merged a while ago :-) closing.

py4j-java/src/test/java/py4j/examples/ReturnerExample.java
```diff
@@ -46,7 +46,7 @@ public int computeInt(IReturnConverter returner) {
 	public int computeNothing(IReturnConverter returner) {
 		return 1;
 	}
-
+
```
fabiencelier: useless new whitespace

scottslewis: This pr has been merged.   Feel free to create pr to remove any unnecessary whitespace (in this pr or elsewhere).

Issue #137, opened 2014-06-07
Provide python stack trace on Python Error


bartdag:
As reported by a user of Py4J, exceptions thrown by Python (e.g., in a Python callback) are swallowed by a Py4JException: they are correctly reported, but the trace is lost.

One solution would be to use the same protocol that we have on the Python Side: if Java throws an error, it can include the stack trace after the "x" (protocol token indicating that an exception occurred).

sangeeta28:
Hi, I have a very basic pyspark code running on an EMR cluster. I am using log4j for logging. I got my logger from the sparkcontext using logger = sc._jvm.org.apache.log4j.Logger.getLogger(__name__)
But, I am not unable to log an exception stacktrace to this logfile. Is there a way to log everything to a log file?

torokati44:
I'd like to just quietly signal my interest in seeing this issue resolved. It would make Py4J so much more useful!

scottslewis:
>I'd like to just quietly signal my interest in seeing this issue resolved. It would make Py4J so much more >useful!

+1.   FWIW Barthelemy, now the osgiservicebridge makes it relatively easy to expose multiple python objects as osgi service instances this error display is more obviously a problem...since the python impl of any of the instances can throw...and just gets back to java as a generic protocol error...i.e. nothing to tell what went wrong.

PR #294, opened 2017-11-06
Provide more information if a Python callback throws (#137)

batterseapower:
I've taken the simple approach of sending Py4JJavaExceptions back to the Java side as references, and everything else as a string. This is a big improvement on just being told that *something* has gone wrong (and perhaps being able to dig out extra info in the logging).

Disclaimer: I couldn't run all the tests -- they just hang on my machine. Possibly related to some missing cleanup code in the tests -- I can see quite a few places where we might fail to shut down some process and then wait indefinitely for it to terminate. Maybe your CI will be able to run these for me?

My two new tests *do* pass.


py4j-python/src/py4j/clientserver.py
```diff
@@ -521,18 +522,27 @@ def wait_for_commands(self):
...
+            params = self._get_params(input)
+            return_value = getattr(self.pool[obj_id], method)(*params)
+            return proto.RETURN_MESSAGE + proto.SUCCESS +\
+                get_command_part(return_value, self.pool)
+        except Exception as e:
+            logger.exception("There was an exception while executing the "
+                             "Python Proxy on the Python Side.")
+
+            if isinstance(e, proto.Py4JJavaError):
+                java_exception = e.java_exception
```
bartdag: Just to make sure I understand, wouldn't this hide the python proxy?

example:

Java.m1() calls Python.p1()
Python.p1() calls Java.m2()
Java.m2() raises FooBarException
Python.p1() raises a Py4JJavaError(value=FooBarException)
Java.m1() receives the FooBarException instance

m1() has almost no clue that it called p1() (there is probably a mention to a proxy in m2() stack trace).

py4j-java/src/main/java/py4j/Protocol.java
```diff
@@ -449,16 +449,17 @@ public final static Object getReference(String commandPart, Gateway gateway) {
...
 
-	public final static Object getReturnValue(String returnMessage, Gateway gateway) {
-		Object returnValue = null;
-
+	public final static Object getReturnValue(String returnMessage, Gateway gateway) throws Throwable {
+		final Object result = getObject(returnMessage.substring(1), gateway);
 		if (isError(returnMessage)) {
-			throw new Py4JException("An exception was raised by the Python Proxy. Return Message: " + returnMessage);
+			if (result instanceof Throwable) {
+				throw (Throwable) result;
```
bartdag: I think I would prefer to chain the exception: `throw new Py4JException("An exception...", result)` so the caller knows that something involving a Python proxy went wrong. Even if the exception originated from the Java code, it still went through a Python proxy.

bartdag:
Thanks for your pull request and for writing tests!

My original idea was to replicate what we are doing on the Java side: save the exception reference in the Python objects pool and send back the exception reference to the Java side. The problem is that we would need to create some sort of ProxyException interface because it does not make sense to send back a Python reference that does not implement a Java interface. Maybe I can leave that to the elusive 0.11/1.0 version and keep this workaround for now considering how many users have requested that feature!

I'll have to test the code locally a bit more and I need to think about the way you re-throw java exceptions to make sure it would not break existing code. Hopefully, I will merge this in the coming week!

Thanks again!

py4j-java/src/main/java/py4j/Protocol.java
bartdag: Yeah, thinking more about it, `throw (Throwable) result` would break backward compatibility because user code currently only expects Py4JException and not any other throwable.

batterseapower: Fair enough. The user can always do this unwrapping themselves if they wish to.

py4j-python/src/py4j/clientserver.py
batterseapower: Yeah it does hide p1 from the stack trace, which is an unfortunate side-effect.

What I'm trying to support here is allowing Python to implement Java interfaces more usefully. e.g. let's say there is a interface:

  interface Stream {
    Object getNext() throws NoSuchElementException;
  }

Where NoSuchElementException is being used to signal the end of the stream. Right now there is no way to signal the stream end if you implement this interface in Py4J because all we can throw is a Py4JException.

py4j-java/src/main/java/py4j/Protocol.java
bartdag: Thanks for the explanation, I wanted to do something similar at one point but got lost in a more convoluted solution. Yours is definitively simpler and more elegant.

Backward compatibility is extremely important for Py4J users (we could literally break more than 100k user programs) so could you:

1. add a parameter in CallbackServerParameters/PythonParameters: propagateJavaException=False (by default we keep the former behavior and only add the error message).
2. before sending back the java exception, check the value of the parameter.
3. add yourself to the authors text file :-)

The parameter will eventually be True by default (for 0.11 or 1.0 release).

Thanks!

batterseapower: Sure thing: I've added this flag. It does make sense to provide an escape hatch people can use to get back the old behaviour if they really need it.

Thanks for your feedback!

scottslewis:
Hi batterseapower and Barthelemy,

I like this contribution, but have a suggestion:   Could it be easily enhanced to supply the python stack trace back to the java Exception?   Python traceback has format_exeception method, which if communicated to the java side in the ERROR message could easily be parsed in java and passed to Throwable.setStackTrace.

bartdag:
Hi Scott,

if the error is caused by a Python error (as opposed to a Java exception), the stack trace is sent to the Java side:

https://github.com/bartdag/py4j/pull/294/files#diff-8716002e9217465f36c981ff28019e09R550

If I understand correctly, you would parse the string (split by line breaks and other syntactic markers) and create an array of StackTraceElement?

batterseapower:
I don't feel particularly motivated to implement this, because I only use the stack trace for debugging, and in that case the string representation is perfectly adequate. Personally I would actually find it a bit surprising to see a Java exception whose StackTraceElement[] actually somehow encoded a Python stack trace: my expectation is that a Java exception contains a Java stack trace.

If you did want to encode the Python stack trace as a Java one you would have to make some "interesting" choices about how to do this mapping -- e.g. how do fill in the "declaring class" field of the Java type if the Python exception was not thrown from a class?

Furthermore, you probably wouldn't want to setStackTrace on the Py4JException itself, because then you'd use lose the Java stack trace of the Py4JException itself. So you'd probably want to set it on a different exception and then set that exception as a "cause" of the Py4JException. So it's all starting to look a bit awkward...

If we did have this feature I think it should be done by ignoring StackTraceElement and instead having a getPythonStackTrace() method on the PyJ4Exception, returning our own data type that can faithfully represent Python stack traces. But like I say, I don't have a use for this myself.

bartdag:
Thanks for your comment @batterseapower . I should have been clearer too: I do not think this pull request needs to be further modified unless the string representation is a blocker for future features and I am not inclined to translate Python stack trace into Java stack trace, but I wanted to make sure I understood Scott's idea correctly.

scottslewis:
howdy,

I think there are 4 options for the python stack trace:

1) Don't include it at all
2) Include the stack trace as a string in the exception message
3) Put it into the stack trace of the Py4jException or a new exception that is cause of Py4jException
4) Create a exception class method: String[] getPythonStackTrace()

I think what you are doing now is 2, is that right?    I don't think 2 is a great idea, because it mixes the exception message and stack.   1 is worse because you lose the stack trace totally.

The reason why I think there is incentive to reuse java's stack trace element is that there's lots of java tooling (e.g. Eclipse, Intellij, etc) that knows how to deal with/present stack traces with arbitrary nesting of causes, that would have no idea about getPythonStackTrace().   

"...e.g. how do fill in the "declaring class" field of the Java type if the Python exception was not thrown from a class?"  I agree this is a little awkward given the differences between java and python, but my inclination would be that a constant class name ...e.g. "Py4jPython" would not be too upsetting.  And I think traceback.extract_tb() would put things into a form that could work with new StackTraceElement().

"So you'd probably want to set it on a different exception and then set that exception as a "cause" of the Py4JException. "    I agree with this thought...i.e. I think it makes sense to have a Py4jException caused by a PythonException as nesting in this manner is common.

Another approach would be to create a PythonException class that takes the python stack trace as a String and overrides Throwable.printStackTrace() by calling getPythonStackTrace().   That would be much easier, and make it unnecessary to fuss with setStackTrace, at least in the immediate term.   For example:

public class PythonException extends Exception {

	private String pythonStackTrace;
	
	public PythonException(String pythonMessage, String pythonStackTrace) {
		super(pythonMessage);
		this.pythonStackTrace = pythonStackTrace;
	}

	public String getPythonStackTrace() {
		return this.pythonStackTrace;
	}
	
	@Override
	public void printStackTrace(PrintStream ps) {
		ps.println("python message: "+getMessage());
		ps.println(getPythonStackTrace());
	}
	@Override
	public void printStackTrace(PrintWriter pw) {
		pw.println("python message: "+getMessage());
		pw.println(getPythonStackTrace());
	}
}

This isn't quite all there is to it, but pretty close.

jonahkichwacoders:
+1 on just using formatted python traceback in message of java exception. 

_On my previous project (AnalysisRPC that is now part of Eclipse Triquetrum) I did all the work of making stack trace elements, etc. It meant that things like the JUnit view in Eclipse displayed each frame in the python stack trace properly, but in the end unless you know someone needs to programatically analyse the stack trace, I found it wasn't really worth it. The other benefit is that a call back and forth (even multiple times) from Java to Python would create a stack trace that showed the user perspective of the stack trace by stitching the Java and Python stack traces together, but omitting the RPC layer. This meant the stack traces would have only the user code in it._

_For reference you can look up [Triquetrum](https://github.com/eclipse/triquetrum/tree/b27ff5c8a9b549333aef70142272ceae9ba4aac2/plugins/extras/org.eclipse.triquetrum.scisoft.analysis.rpc/) or ping me._

bartdag:
Thanks everyone for your input.

Considering that I watched yesterday a talk on Youtube that at some point complained about missing callback error information in Py4J* and that I regularly see tweets, questions, and github issues on this, I will go ahead and merge this pull request.

There are many things I would like to do with exceptions such as raising Java exceptions (NoSuchElementException) from Python code, chaining Python and Java exceptions together (now possible with Python 3, but wasn't possible in Python 2), catching specific type of Java exceptions in Python code, converting specific types of Java exceptions in Python, etc. The last time I tried though, I encountered fundamental issues related to the mismatch between the type system and the exception hierarchy in Java and Python. 

I believe that converting Python stack trace to Java StackTraceElement would need to be done in the larger context of improving interoperability between Java and Python exceptions.

*It probably took a longer time to complain, prepare and deliver the talk than the time batterseapower took to fix this.

scottslewis:
Hi Barthelemy,

Where is the python stack trace info going to be then?   If it's in the message for this release and elsewhere for future releases, it will create a new backward compatibility problem for Py4j consumers.

scottslewis:
Barthelemy and jonahkichwacoders,

I agree that stitching python stack traces into stack trace elements is too much...at least for python 2.  But I don't think that implies the python stack trace should be embedded in the message string.   

The use of a PythonException class however, as per previous comment, would allow future efforts without introducing backward compatibility problems.

bartdag:
Hi Scott, I'm not sure I understand what you mean by the stack trace in the exception message being backward incompatible? Can you explain further?

I do get your point that PythonException would be cleaner, but I'm not willing to introduce a new type of exception in the short term without a clearer idea of how the ideal handling of exceptions on both sides would look like (this could certainly be discussed in a new issue).

scottslewis:
Hi All,

What I mean by the exception message being backward incompatible is that if the stack trace is embedded in the message, to remove it in the future could require Py4j consumers to change their code.  

For example, there are plenty of situations (on java side) where only the message exception should be presented...e.g. via some UI...and Throwable.getMessage()/getLocalizedMessage() allows this to be done.   If the python stack trace were included in the message, then it would be necessary to parse what's returned...e.g. by line break...to present just the message.   Doing so, however, means that if the stack trace is later moved out of the message, that backward compatibility is broken.

>I do get your point that PythonException would be cleaner, but I'm not willing to introduce a new type >of exception in the short term without a clearer idea of how the ideal handling of exceptions on both >sides would look like (this could certainly be discussed in a new issue).

I can't honestly say how it should look like...especially given the need to support python 2 and 3. 
  Although it has some appeal, I don't think it makes sense to try to create a java-side hierarchy that matches the python exception/error hierarchy.   That seems unnecessarily complicated and difficult to me...especially because nesting doesn't seem to be a common thing in python, whereas it is common in java.   And python isn't all class-based as batterseapower has pointed out.

I just think it would be useful to hedge against something more than Py4jException to add PythonException and keep the python message and stack trace separate...because on the java-side...at least...the message and stack trace are frequently used differently.   

The thing that I like about PythonException is that it keeps future options open (change of impl and subclasses) and keeps a separation between the python message and stack trace.  And it has a low code size/complexity cost.

bartdag:
Hi Scott, 

I understand your concern about putting everything in the exception message, but at least, in the future, the message will always be present and only the content may change. If we introduce a new class (PythonException), we will have to support it and keep it even though we are not sure this is the right way to go.

I agree that Py4J needs a better story when it comes to exceptions: it should be easier to raise them on both sides, capture them and inspect them programmatically without having to perform black magic on the exception messages. I prefer to introduce new types only when we'll have a better picture of what the whole thing should look like.

@batterseapower I am locally testing your branch and I'm running into weird test issues that I don't see in master. I'll update this pull request when I see what the problem is. I'm also a bit surprised circleci hasn't kicked in on this pull request :-(

bartdag:
@batterseapower I configured an option within CircleCI to build pull requests from forked repositories. You should now see the test results in your next push.

I see a few tests failing:

```
nosetests-2.7 -v --nologcapture --nocapture py4j.tests.java_callback_test:IntegrationTest.testProxyError

nosetests-2.7 -v --nologcapture --nocapture py4j.tests.client_server_test:IntegrationTest.testErrorInPy4J
```

I'll focus on the client_server_test error: this one is always trickier. If you can take a look at testProxyError that would be great, otherwise I'll investigate by the end of the week.

py4j-python/src/py4j/tests/client_server_test.py
```diff
@@ -317,6 +324,56 @@ def testErrorInPy4J(self):
             # Check that the connection is not broken (refs #265)
             val = client_server.jvm.java.lang.Math.abs(-4)
             self.assertEqual(4, val)
+
```
bartdag: @batterseapower small error here: the line `client_server.shutdown()` was removed from testErrorInPy4J and put at the end of testErrorInPythonCallback. Both tests need this line :-)

This explains why I was getting an error with testErrorInPy4j.

scottslewis:
Hi All,

I disagree with your choice Barthelemy...as it is guaranteed to cause some consumers problems when things change in the future.   I accept that you would like to understand how the 'whole story' is going to be but I don't think it's possible to have the whole story at this point, especially with python 2 still fully in the picture.    It is possible...IMHO...to predict that eventually you will have to introduce more java-side classes for representing python exceptions...distinct from java exceptions...and use java's nesting/cause structure to do so.   That's why I think it makes sense to introduce 1 small root PythonException class now so that it can be easily extended in future.

But of course you are the project lead so it is your choice.   I'm only trying to give you the perspective of a grateful consumer with some inter-process comm API experience.

Scott

batterseapower:
Yeah I was hoping CircleCI would kick in because I couldn't get your tests running locally reliably even using master. Good to hear it is fixed now. I probably won't be able to look at this further until December 11th when I'm back from holiday.

batterseapower:
I can reproduce the 'testErrorInPy4J' failure, and adding the shutdown fixed it. But I can't reproduce your other failure:

> $ /c/Python27/Scripts/nosetests-2.7 -v --nologcapture --nocapture py4j.tests.java_callback_test:IntegrationTest.testPro
> xyError
> testProxyError (py4j.tests.java_callback_test.IntegrationTest) ... ok
> 
> ----------------------------------------------------------------------
> Ran 1 test in 1.603s
> 
> OK

This is with JDK 1.8.0_131, python 2.7.13, Windows 10. Works with my Python 3 install too (Python 3.6.1).

Can you give a bit more detail about the problem you see? CircleCI seems to be timing out, and it's not clear to me that this is to do with my changes.

batterseapower:
Rerunning CircleCI worked (no test timeout), and it still doesn't reproduce the testProxyError failure. So I'm not sure what's going on there.

py4j-python/src/py4j/tests/java_array_test.py
```diff
@@ -68,7 +68,7 @@ def testCharArray(self):
         char_class = self.gateway.jvm.char
         char_array = self.gateway.new_array(char_class, 2)
         char_array[0] = "a"
-        self.assertAlmostEqual(char_array[0], "a")
+        self.assertEqual(char_array[0], "a")
```
bartdag: this should work (assertAlmostEqual) even on Python 2.6, but it was probably a copy-paste error. thanks for fixing it.

bartdag:
Thanks @batterseapower this seems to work. Not sure why I was reliably having this error but everything passed on CircleCI.

The two changes that you can make or that I can make are:
1. Rebase and squash everything into one commit (https://www.py4j.org/contributing.html#branch-name-and-commit-message-format)
2. propagate_java_exceptions=False should be the last parameter of each of the Parameter classes to avoid breaking compatibility.

If you're too busy, I can do it too.

batterseapower:
Hope this is suitable.

bartdag:
Looks great, thanks a lot!

batterseapower:
Anything blocking this from being merged?

bartdag:
Merged in aa6c53b59027925a426eb09b58c453de02c21b7c

Thanks!

scottslewis:
I issued provide a pull request 

https://github.com/bartdag/py4j/pull/303

That adds a new class:  PythonThrowable.   This class parses the String from Python traceback.format_exc and allows the presentation of the Python exception info (stack trace, exception type, exc msg) in both the original python form (most recent stack element last) or in a java style (most recent stack element first).   Either form can be used.   Included test code and a small addition to Py4JException class to use PythonThrowable.

Please consider this addition in combination with the fix for this issue.

PR #308, opened 2018-02-24
Adds a note for the fact that __str__ could return non-ascii unicodes in Python 2

HyukjinKwon:
This PR proposes to add a simple note and comment for https://github.com/bartdag/py4j/issues/306. It might be trivial but good to know.


bartdag:
Thanks! Merged!

PR #312, opened 2018-04-18
Add support for auth tokens to gateway and callback connections.

### LINKED ISSUE #268: Security problems
Py4j runs a TCP server, by default on localhost.  This is subject to both local and (potential) remote code execution exploits:

- Local: anyone running on the same machine could control the JVM.  This is a disaster on multi-user systems.
- Remote: DNS rebinding attacks allow one to use client-side Javascript to make arbitrary HTTP requests.  If those are not properly rejected, this could allow remote code execution.

The solution to both of these is to use a secure temporary file to store a cryptographic key, which both client and server use to authenticate all data.  This must be enabled by default.

vanzin:
This change adds authentication to both the gateway and callback
servers, in the form of a shared authentication token.

The generation and distribution of tokens is left for applications
to do. The library just provides the means for enforcing the tokens.
The only exception is when using the gateway server command line,
in which case it will generate a token when requested.

By default, like previously, there is no authentication.

The auth protocol is very, very simple: when auth is enabled, the
servers expect the first line of content from the client connection
to be the auth token. If the token doesn't match, the server will
shut down the connection. Note this means that the token is sent
in plain text over the connection, meaning that for security it
requires the use of a local socket, or SSL for remote sockets.

Fixes #268

### DIFF
--- a/py4j-java/src/main/java/py4j/CallbackConnection.java
+++ b/py4j-java/src/main/java/py4j/CallbackConnection.java
@@ -47,9 +47,9 @@ import javax.net.SocketFactory;
  * Default implementation of the CommunicationChannel interface using TCP
  * sockets.
  * </p>
- * 
+ *
  * @author Barthelemy Dagenais
- * 
+ *
  */
 public class CallbackConnection implements Py4JClientConnection {
 
@@ -74,6 +74,8 @@ public class CallbackConnection implements Py4JClientConnection {
 
 	private final int nonBlockingReadTimeout;
 
+	private final String authToken;
+
 	public CallbackConnection(int port, InetAddress address) {
 		this(port, address, SocketFactory.getDefault());
 	}
@@ -94,6 +96,23 @@ public class CallbackConnection implements Py4JClientConnection {
 	 *            must absolutely be non-blocking.
 	 */
 	public CallbackConnection(int port, InetAddress address, SocketFactory socketFactory, int readTimeout) {
+		this(port, address, socketFactory, readTimeout, null);
+	}
+
+	/**
+	 *
+	 * @param port The port used to connect to the Python side.
+	 * @param address The address used to connect to the Java side.
+	 * @param socketFactory The socket factory used to create a socket (connection) to the Python side.
+	 * @param readTimeout
+	 *            Time in milliseconds (0 = infinite). Once connected to the Python side,
+	 *            if the Java side does not receive a response after this time, the connection with the Python
+	 *            program is closed. If readTimeout = 0, a default readTimeout of 1000 is used for operations that
+	 *            must absolutely be non-blocking.
+	 * @param authToken Token for authenticating with the callback server.
+	 */
+	public CallbackConnection(int port, InetAddress address, SocketFactory socketFactory, int readTimeout,
+			String authToken) {
 		super();
 		this.port = port;
 		this.address = address;
@@ -104,6 +123,7 @@ public class CallbackConnection implements Py4JClientConnection {
 		} else {
 			this.nonBlockingReadTimeout = DEFAULT_NONBLOCKING_SO_TIMEOUT;
 		}
+		this.authToken = authToken;
 	}
 
 	public String sendCommand(String command) {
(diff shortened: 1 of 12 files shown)


vanzin:
@bartdag gentle ping.

vanzin:
Not sure how to retrigger the test without making some dummy change...

py4j-java/src/main/java/py4j/CallbackConnection.java
```diff
@@ -207,6 +227,18 @@ public void start() throws IOException {
 		socket.setSoTimeout(blockingReadTimeout);
 		reader = new BufferedReader(new InputStreamReader(socket.getInputStream(), Charset.forName("UTF-8")));
 		writer = new BufferedWriter(new OutputStreamWriter(socket.getOutputStream(), Charset.forName("UTF-8")));
+
+		if (authToken != null) {
```
bartdag: maybe extract in a private method?

py4j-java/src/main/java/py4j/GatewayConnection.java
```diff
@@ -203,6 +210,19 @@ public void run() {
 		boolean reset = false;
 		Throwable error = null;
 		try {
+			if (authToken != null) {
```
bartdag: same comment as in clientconnection

py4j-java/src/main/java/py4j/Protocol.java
```diff
@@ -331,6 +331,10 @@ public final static String getOutputErrorCommand() {
 		return ERROR_COMMAND;
 	}
 
+	public final static String getOutputSuccessCommand() {
+		return "" + RETURN_MESSAGE + SUCCESS + END_OUTPUT;
```
bartdag: we usually use Protocol.getOutputVoidCommand();

py4j-java/src/main/java/py4j/StringUtil.java
```diff
@@ -33,16 +33,20 @@
...
- * 
+ *
  */
 public class StringUtil {
 
 	public final static char ESCAPE_CHAR = '\\';
 
 	public static String escape(String original) {
-		return original.replace("\\", "\\\\").replace("\r", "\\r").replace("\n", "\\n");
+		if (original != null) {
```
bartdag: why is it necessary to add a check here considering that unescape has no such check?

py4j-python/src/py4j/protocol.py
```diff
@@ -196,11 +199,14 @@ def unescape_new_line(escaped):
 
     :rtype: the original string
     """
-    return ESCAPE_CHAR.join(
-        "\n".join(
-            ("\r".join(p.split(ESCAPE_CHAR + "r")))
-            .split(ESCAPE_CHAR + "n"))
-        for p in escaped.split(ESCAPE_CHAR + ESCAPE_CHAR))
+    if escaped:
```
bartdag: same comment as in StringUtil

py4j-java/src/main/java/py4j/StringUtil.java
vanzin: I only use `escape` for the auth token, since it's a string and someone might add a new line there and break things. I don't need unescape for it.

This avoids adding another method to do a null-safe escape.

py4j-python/src/py4j/protocol.py
vanzin: Same reason.

vanzin:
@bartdag I think I addressed all your comments.

I left the "authentication as a command" changes in a separate commit since I'm not so sure that's a useful change. In my view, it makes the code more complicated and abuses the command abstraction, and doesn't really seem to make things any better overall. But your call.

bartdag:
Thanks for your latest commits, they do address all my concerns. I will probably have a shot at refactoring the way the auth command is initialized/called, but it makes the communication easier to debug (especially with the error message you provided). I was expecting the contribution over the weekend or early next week and had planned to work on the release during the April 28th weekend. I'll see what I can do during the weekend to at least make sure that everything is ready for a release.

Thanks a lot for your contribution and for taking the time to make your changes backward compatible: that means a lot to me and to Py4J users.

bartdag:
Just a heads up about the integration of this patch: at this point, I'm happy with the patch and the small refactoring I did. While testing though, I noticed that errors related to a missing token or a wrong token are not 100% obvious if logging is turned off so I'll need to spend a bit more time tomorrow trying to make it more obvious. I get many many support requests for simple errors such as a wrong ip address or a wrong port, so I am confident that wrong tokens will significantly increase my support load if it's not more obvious.

bartdag:
I will also fix the copyright year. There is a license template file that needs to be changed and spotlessApply will fix all copyright headers.

PR #353, opened 2019-03-23
Add support for specifying current working directory of the Java process

Chraneco:
Thank you very much for providing this library!

For our use case we would need to be able to specify the current working directory of the launched Java process. This pull request adds this feature.


bartdag:
Thanks and apologies for the delay.

Merged in master.

PR #366, opened 2019-07-03
Partially resolve #320 by enabling user code to shut down a Java process manually

### LINKED ISSUE #320: When starting new Java server-processes in a loop, the previous one does not get stopped
I am experiencing difficulties when starting continuously new Java server processes within a loop. See below for a complete basic example.

If using `launch_gateway(die_on_exit=False, ...)`,  the previous process gets closed correctly by `gateway.shutdown()`. However in that case the Java server-process will not be closed if the parent Python-process is closed with `Crtl + C` on a Linux command line - this is also not acceptable in my case.

That means I am forced to use `die_on_exit=True`. But why does `gateway.shutdown()` does not stop the Java server-process?

Can you please comment on my usage of Py4J. Do I use it in the wrong way? Is there a work-around to ensure that creating the Java server-processes safely stops the previous one and at the same time the Java server-process is always stopped if the Python process dies?

**Python code:**
```
from py4j.java_gateway import JavaGateway, GatewayParameters, launch_gateway


def start_api_server():
    port = launch_gateway(classpath="C:\\Users\\User\\Documents\\code\\py4jjavademo\\target\\classes",
                          die_on_exit=True)

    global gateway
    gateway = JavaGateway(gateway_parameters=GatewayParameters(port=port))
    string_generator = gateway.jvm.com.demo.EntryPoint().getStringGenerator()

    return string_generator


def stop_api_server():
    global gateway
    gateway.shutdown()


if __name__ == '__main__':
    for i in range(100):
        try:
            string_generator = start_api_server()

            print(string_generator.get("Content"))
        finally:
            stop_api_server()
```

**Java code:**
```
package com.demo;


public class EntryPoint {

   private StringGenerator stringGenerator;

   public EntryPoint() {
      stringGenerator = new StringGenerator();
   }

   public StringGenerator getStringGenerator() {
      return stringGenerator;
   }
}
```

```
package com.demo;


public class StringGenerator {

    public String get(String content) {
        return "A String: " + content;
    }
}
```

dHannasch:
This copies @bartdag's code from https://github.com/bartdag/py4j/issues/320 so that the Popen object for the Java process is saved on the JavaGateway, and can thus be referenced later, including to terminate it by sending the end-of-line character it is waiting for.

As far as I can tell, as of this writing Py4J has no native support for the use-case where a Java process launched from Python needs to die if the launching Python process dies *or* if deliberately terminated by the still-running Python process.

As a quick reminder of what we're talking about with the Java process not being shut-down-able when `die_on_exit=True`:

```
import py4j.java_gateway
gateway = py4j.java_gateway.JavaGateway.launch_gateway(die_on_exit=True)
gateway.close_callback_server(True)
gateway.close(False, True)
gateway.shutdown_callback_server(True)
gateway.shutdown(True)
# Do other things...
# As shown by e.g. ps aux, the Java process is still running!
```

@bartdag also mentioned preferring to add an argument `terminate_launched_jvm` to `shutdown()` for a better solution, but I wasn't sure how that was intended to work, so I thought I'd first put this forward to fish for interest in at least getting this much in. Just this already enables shutting down the Java process from user Python code, purely by virtue of saving a reference to the Java process.

Personally, I think it would be even better to simply make a Popen reference a standard feature of the JavaGateway, just like the `java_gateway_server`:

    * The `java_process` field of a `JavaGateway` instance is a
      subprocess.Popen object for the Java process that the `JavaGateway`
      is connected to, or None if the `JavaGateway` connected to a preexisting
      Java process.

I don't know if that would be too confusing that a `JavaGateway` might have no `java_process` if it is connected but was connected to a preexisting Java process. But it's certainly great in the case where the Java process *was* launched from Python. Having access to the Popen object is just endlessly useful. (It would be nice to have even if the `JavaGateway` did connect to a preexisting Java process, but as far as I know in that case it's impossible.) This is a trade-off between convenience and potential confusion, much like `JavaGateway` redirecting method calls to the `entry_point`.

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -220,7 +220,7 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
                    die_on_exit=False, redirect_stdout=None,
                    redirect_stderr=None, daemonize_redirect=True,
                    java_path="java", create_new_process_group=False,
-                   enable_auth=False, cwd=None):
+                   enable_auth=False, cwd=None, return_proc=False):
     """Launch a `Gateway` in a new Java process.
 
     The redirect parameters accept file-like objects, Queue, or deque. When
@@ -270,6 +270,8 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
         authentication token when connecting.
     :param cwd: If not None, path that will be used as the current working
         directory of the Java process.
+    :param return_proc: If True, returns the Popen object returned when the JVM
+        process was created.
 
     :rtype: the port number of the `Gateway` server or, when auth enabled,
             a 2-tuple with the port number and the auth token.
@@ -348,9 +350,17 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
         quiet_close(stderr)
 
     if enable_auth:
-        return (_port, _auth_token)
+        output = (_port, _auth_token)
     else:
-        return _port
+        output = _port
+
+    if return_proc:
+        if isinstance(output, tuple):
+            output = output + (proc, )
+        else:
+            output = (_port, proc)
+
+    return output
 
 
 def get_field(java_object, field_name):
@@ -1715,6 +1725,12 @@ class JavaGateway(object):
     * The `jvm` field of `JavaGateway` enables user to access classes, static
       members (fields and methods) and call constructors.
 
+    * The `java_process` field of a `JavaGateway` instance is a
+      subprocess.Popen object for the Java process that the `JavaGateway`
+      is connected to, or None if the `JavaGateway` connected to a preexisting
+      Java process (in which case we cannot directly access that process from
+      Python).
+
     Methods that are not defined by `JavaGateway` are always redirected to
     `entry_point`. For example, ``gateway.doThat()`` is equivalent to
     ``gateway.entry_point.doThat()``. This is a trade-off between convenience
@@ -1726,7 +1742,8 @@ class JavaGateway(object):
             python_proxy_port=DEFAULT_PYTHON_PROXY_PORT,
             start_callback_server=False, auto_convert=False, eager_load=False,
             gateway_parameters=None, callback_server_parameters=None,
-            python_server_entry_point=None):
+            python_server_entry_point=None,
+            java_process=None):
         """
         :param gateway_parameters: An instance of `GatewayParameters` used to
             configure the various options of the gateway.
@@ -1738,6 +1755,9 @@ class JavaGateway(object):
 
         :param python_server_entry_point: can be requested by the Java side if
             Java is driving the communication.
+
+        :param java_process: the subprocess.Popen object for the Java process
+            that the `JavaGateway` shall connect to, if available.
         """
 
         self.gateway_parameters = gateway_parameters
@@ -1791,6 +1811,8 @@ class JavaGateway(object):
             self._eager_load()
         if self.callback_server_parameters.eager_load:
             self.start_callback_server(self.callback_server_parameters)
+
+        self.java_process = java_process
 
     def _create_gateway_client(self):
         gateway_client = GatewayClient(
@@ -2114,14 +2136,15 @@ class JavaGateway(object):
             redirect_stdout=redirect_stdout, redirect_stderr=redirect_stderr,
             daemonize_redirect=daemonize_redirect, java_path=java_path,
             create_new_process_group=create_new_process_group,
-            enable_auth=enable_auth, cwd=cwd)
+            enable_auth=enable_auth, cwd=cwd, return_proc=True)
         if enable_auth:
-            _port, _auth_token = _ret
+            _port, _auth_token, proc = _ret
         else:
-            _port, _auth_token = _ret, None
+            _port, proc, _auth_token = _ret + (None, )
         gateway = JavaGateway(
             gateway_parameters=GatewayParameters(port=_port,
-                                                 auth_token=_auth_token))
+                                                 auth_token=_auth_token),
+            java_process=proc)
         return gateway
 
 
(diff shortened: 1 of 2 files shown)


dHannasch:
Hmm. The CI build is complaining about this:

* What went wrong:
A problem occurred configuring root project 'py4j-java'.
> java.lang.NoSuchMethodError: org.gradle.api.internal.file.UnionFileCollection.add(Lorg/gradle/api/file/FileCollection;)Lorg/gradle/api/file/FileCollection;

Does anyone know what could cause that? As you can see, I didn't touch any files in 'py4j-java', only in 'py4j-python'.

bartdag:
I encountered that locally. I think i need to change the ci script to use the gradle wrapper and not the global gradle.

I also need to update the gradle version:

```diff
diff --git a/py4j-java/gradle/wrapper/gradle-wrapper.properties b/py4j-java/gradle/wrapper/gradle-wrapper.properties
index 8142059..9d94a16 100644
--- a/py4j-java/gradle/wrapper/gradle-wrapper.properties
+++ b/py4j-java/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.11-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
```

```diff
diff --git a/.travis.yml b/.travis.yml
index 5f6829b..02f260a 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,11 +54,11 @@ install:
   #- if [ $TEST_SUITE == "java6" ] ; then pyenv install 2.7.12; pyenv global 2.7.12; pip install --user nose; fi
   - cd py4j-java
   # Useful in case the build stops working because of version issues.
-  - gradle --version
+  - ./gradlew --version
   - echo `java -version` && echo `javac -version` && echo `python --version`
   - echo $PATH
   - echo $JAVA_HOME
-  - gradle clean
+  - ./gradlew clean
 before_script:
   # Enable IPv6
   - echo 0 | sudo tee /proc/sys/net/ipv6/conf/all/disable_ipv6
@@ -66,8 +66,8 @@ script:
   # Our gradle build file will take this env variable and set the bootstrap
   # class path to JDK 6 because source compatibility is set to 6.
   - if [ $TEST_SUITE != "java8-3.7" ] ; then export JDK6_BOOT_CLASSPATH=/usr/lib/jvm/java-6-openjdk-amd64/jre/lib/rt.jar:/usr/lib/jvm/java-6-openjdk-amd64/jre/lib/jsse.jar; fi
-  - gradle check
-  - gradle assemble
+  - ./gradlew check
+  - ./gradlew assemble
   - cd ../py4j-python
   # Only run flake8 once to save time.
   - if [ $TEST_SUITE == "java8" ] ; then flake8; fi
```

If you want to try it in your PR, go ahead. I should have time to look at this during the weekend only.

py4j-python/src/py4j/java_gateway.py
```diff
@@ -2110,14 +2120,16 @@ def launch_gateway(
...
-            _port, _auth_token = _ret
+            _port, _auth_token, proc = _ret
         else:
-            _port, _auth_token = _ret, None
+            _port, proc, _auth_token = _ret + (None, )
         gateway = JavaGateway(
             gateway_parameters=GatewayParameters(port=_port,
                                                  auth_token=_auth_token))
+        # NOTE 3: the Popen object is now available with gateway._proc
+        gateway._proc = proc
```
bartdag: could you add this member to JavaGateway.__init__ initialized to None so that even if the gateway is not launched through launch_gateway, the member is defined?

bartdag:
Thanks for this pull request! A test showing a simple usage scenario of what can be done with the POpen object would be nice too.

dHannasch:
Hmm. The Python 3.6 build (and only the Python 3.6 build) fails on testPythonToJavaToPythonClose:

```
testPythonToJavaToPythonClose (py4j.tests.memory_leak_test.ClientServerTest) ... 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

nose seems to have the wrong capitalization there, it should be https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received . Unfortunately nothing at https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received looks particularly promising.

I didn't touch anything to do with memory_leak_test.ClientServerTest, as far as I can tell. There are other tests in memory_leak_test that use the JavaGateway, but ClientServerTest does not. On the other hand, I don't know why a build problem would suddenly manifest on testPythonToJavaToPythonClose (py4j.tests.memory_leak_test.ClientServerTest) and not on testPythonToJavaToPython (py4j.tests.memory_leak_test.ClientServerTest) which successfully completed right before it.

Any ideas?

py4j-python/src/py4j/java_gateway.py
dHannasch: Sounds good. I'm calling the new the `JavaGateway` member `java_process` by analogy with `java_gateway_server`, if that sounds like a good name.

dHannasch:
> Thanks for this pull request! A test showing a simple usage scenario of what can be done with the POpen object would be nice too.

Sounds good. I added a test that sends the end-of-line character and verifies that the Java process stops.

Unfortunately, while the code to *do* it is the same between Python 2 and Python 3, the code to *check* that the Java process has stopped requires an ugly if-check. (Well, technically the Python 2 version would probably work for Python 3 as well, but the Python 3 version runs much faster since `wait(2**-5)` returns as soon as the Java process terminates while still being guaranteed not to hang forever. In Python 2 we had to wait a while and then check whether the subprocess had terminated...which, if we waited long enough to be sure it terminated, usually meant we waited too long. So this test will slow down your Python 2 CI build by another quarter of a second. Not a huge amount, but there's no need to slow down the Python 3 CI builds as well.)

PR #381, opened 2019-10-16
JavaGateway.help: Fix deprecation warning in py3.6

AbdealiJK:
In py3.6, a backslash-character pair that is not a valid escape
sequence now generates a DeprecationWarning. This will later be
converted to a SyntaxError.

A docstring is a valid string in python - and in the docstring,
"\*" was present in JavaGateway. Escape it correctly as \\*

Fixes https://github.com/bartdag/py4j/issues/377


bartdag:
Thanks for your pull request. It's now merged in master.

PR #389, opened 2020-01-31
Allow java gateway class to be customized

helenyugithub:


helenyugithub:
Hi @bartdag, do you think this would be a reasonable feature to add? I'd like to support launching my own custom java gateway

py4j-python/src/py4j/java_gateway.py
```diff
@@ -221,7 +221,8 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
                    die_on_exit=False, redirect_stdout=None,
                    redirect_stderr=None, daemonize_redirect=True,
                    java_path="java", create_new_process_group=False,
-                   enable_auth=False, cwd=None, return_proc=False):
+                   enable_auth=False, cwd=None, return_proc=False,
+                   gateway_class="py4j.GatewayServer"):
```
bartdag, line 225: this should be called entry_point_class. Strictly speaking, we are specifying the Application's Entry Point

py4j-python/src/py4j/java_gateway.py
```diff
@@ -273,6 +274,7 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
         directory of the Java process.
     :param return_proc: If True, returns the Popen object returned when the JVM
         process was created.
+    :param gateway_class: the Java Gateway to launch
```
bartdag, line 277:
```
the application's entry point class to launch. It should accept the options --die-on-broken-pipe and --enable-auth. It should write on stdout the port and optionally the auth token.
```

PR #398, opened 2020-04-27
fix #397 : do not ignore port arg anymore

### LINKED ISSUE #397: GatewayServer#main doesn't pick up port argument
I have been experimenting with calling `JavaGateway.launch_gateway` within my Python code as a cheap and cheerful way of getting the Java side up and running.

This worked just fine, until I tried to have two instances running at once, at which point the call to `launch_gateway` seemed to hang indefinitely. I guessed it had to do with a port conflict, but the same thing happened even if I was careful to explicitly specify different ports for the two instances.

My dive into the Py4J source code brought me to the `main` method of class `py4j.GatewayServer` on the Java side, since this is what the Python-land `launch_gateway` method causes to be invoked when it starts up a new JVM process. It looks like this `main` method is supposed to be picking up the port as as final command-line argument, but it is not properly doing so, since the loop over the `args` array seems to stop one element too soon. It looks as though perhaps the logic around getting the port number here was changed with a recent revision, and this loop bound may have been overlooked.

yannrichet:


bartdag:
Wow. I feel bad I let this bug being introduced in a previous pull request.

bartdag:
Merged. Thanks!

PR #419, opened 2020-12-02
Exposing shell parameter in Popen inside launch_gateway

hlasu:
When running our python api from matlab, a console window pops up when Popen is run inside launch_gateway
If instead Popen is run with shell=True, no window pops up.

Therefore we would like to make it possible to run the java process like that in launch_gateway.

This fix does not alter the default behavior, which is to run Popen without shell argument.

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -221,7 +221,7 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
                    die_on_exit=False, redirect_stdout=None,
                    redirect_stderr=None, daemonize_redirect=True,
                    java_path="java", create_new_process_group=False,
-                   enable_auth=False, cwd=None, return_proc=False):
+                   enable_auth=False, cwd=None, return_proc=False, use_shell=True):
     """Launch a `Gateway` in a new Java process.
 
     The redirect parameters accept file-like objects, Queue, or deque. When
@@ -273,6 +273,7 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
         directory of the Java process.
     :param return_proc: If True, returns the Popen object returned when the JVM
         process was created.
+    :param use_shell – If true, Popen will be executed with shell=use_shell
 
     :rtype: the port number of the `Gateway` server or, when auth enabled,
             a 2-tuple with the port number and the auth token.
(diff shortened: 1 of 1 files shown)


py4j-python/src/py4j/java_gateway.py
```diff
@@ -324,8 +325,13 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
     if create_new_process_group:
         popen_kwargs.update(get_create_new_process_group_kwargs())
 
-    proc = Popen(command, stdout=PIPE, stdin=PIPE, stderr=stderr,
-                 cwd=cwd, **popen_kwargs)
+    if use_shell:
```
bartdag, line 328: Hi @hlasu, could we do this instead:

```python
popen_kwargs["shell"] = use_shell
```

py4j-python/src/py4j/java_gateway.py
```diff
@@ -273,6 +273,7 @@ def launch_gateway(port=0, jarpath="", classpath="", javaopts=[],
         directory of the Java process.
     :param return_proc: If True, returns the Popen object returned when the JVM
         process was created.
+    :param use_shell – If True, Popen will be start the java process with shell=True
```
bartdag, line 276: this parameter should be documented and added to the launch_gateway classmethod in the JavaGateway class

bartdag (changes requested):
Thank you for this addition! There are two small changes to make and then we are good to go.

py4j-python/src/py4j/java_gateway.py
hlasu: I have updated code to use 
    popen_kwargs["shell"] = use_shell 
instead as you wanted (the reason I did it like it did before was that i did not want to cause any strange side-effects when using the default value)

I have also changed the class method.

bartdag: Thank you. I also hesitated, but shell=False is supposed to be the default value on all OSes so hopefully all python interpreters follow the API documentation :crossed_fingers:

bartdag (approved):
Thank you. I will merge this during the weekend and hopefully make a release next weekend.

hlasu:
I wonder if there are any news about release-date?

bartdag:
@hlasu I just released 0.10.9.2 on all platforms (sonatype, pypi, and bintray). Thanks for your patience!

hlasu:
np

Thanks for letting me contribute.

cya

Issue #441, opened 2021-10-15
java_tls_test suite fails with test error then hangs indefinitely


JoshRosen:
The `src/py4j/tests/java_tls_test.py` test errors out and then hangs indefinitely, both in local tests and in GitHub Actions.  This problem appears to be specific to that suite, since all other test suites run correctly.

I'm filing this ticket so that I can reference it in a PR porting the tests to GitHub Actions: for now, I'm adding a flag to skip this suite until we can fix it but I don't want the test port to block on that.

When running

```
python -m nose src/py4j/tests/java_tls_test.py
```

I don't see any streaming output. When I interrupt the hanging test with ctrl-C, I see error messages about being unable to connect to the Java server:

```
======================================================================
ERROR: testMethodConstructor (py4j.tests.java_tls_test.TestIntegration)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/joshrosen/py4j/py4j-python/src/py4j/tests/java_tls_test.py", line 90, in testMethodConstructor
    oe1 = self.gateway.jvm.py4j.examples.OperatorExample()
  File "/Users/joshrosen/py4j/py4j-python/src/py4j/java_gateway.py", line 1700, in __getattr__
    "\n" + proto.END_COMMAND_PART)
  File "/Users/joshrosen/py4j/py4j-python/src/py4j/java_gateway.py", line 1036, in send_command
    connection = self._get_connection()
  File "/Users/joshrosen/py4j/py4j-python/src/py4j/java_gateway.py", line 984, in _get_connection
    connection = self._create_connection()
  File "/Users/joshrosen/py4j/py4j-python/src/py4j/java_gateway.py", line 990, in _create_connection
    connection.start()
  File "/Users/joshrosen/py4j/py4j-python/src/py4j/java_gateway.py", line 1132, in start
    raise Py4JNetworkError(msg, e)
Py4JNetworkError: An error occurred while trying to connect to the Java server (localhost:25333)
-------------------- >> begin captured logging << --------------------
py4j.java_gateway: DEBUG: set_default_callback_accept_timeout is deprecated and will be removed in version 1.0. Use CallbackServerParameters instead.
py4j.java_gateway: DEBUG: set_default_callback_accept_timeout is deprecated and will be removed in version 1.0. Use CallbackServerParameters instead.
py4j.java_gateway: INFO: Callback Server Starting
py4j.java_gateway: INFO: Socket listening on ('127.0.0.1', 25334)
[...]

PR #442, opened 2021-10-15
Switch from Travis CI to GitHub Actions

JoshRosen:
This PR switches Py4J's CI from Travis CI to GitHub Actions.

Like the existing Travis build, it runs Py4J's Python and Java test suites using Python 2.7, 3.6, and 3.7 with Java 8. In followup work we can consider expanding the build matrix to cover other Java versions.

The `java_tls_test` suite was failing, so I configured the CI to skip that suite and filed #441 to track fixing and re-enabling it.


JoshRosen:
You can view sample workflow output from my fork at https://github.com/JoshRosen/py4j/actions/runs/1347235622

In order to view output in this PR, you'll have to approve my workflow run: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Issue #403, opened 2020-06-24
Support for Java 11+


pwittchen:
Hi,

It seems that py4j is not working properly with Java 11. I have no issues with Java 8. Do you have plans to support Java versions higher than 8? Inside build configuration, I cannot see details related to Java higher than 8.

Regards,
Piotr

pwittchen:
Can you give any comments on this?

deb75:
It works for me with java 11, cannot set for 12, 13, ...

What kind of bugs did youy experience ?

johann-petrak:
I would also be very interested in the exact details about which Java versions can be used and where problems are to be expected. 

If there are know issues with "recent" Java versions, especially LTS versions, I think this should get mentioned prominently in the README and on the main documentation page.

bartdag:
A detailed error report with stack trace, error message, version of Java, Python, and OS would help tremendously. 

Java 11 is not included in the test suite, but I could try to add it once Travis-ci.org moves to to Travis-ci.com.

There were some changes to the security model of reflection in Java 9 so some code that used to work well in Java 8 may require additional flags (e.g., --illegal-access=warn). There were also some changes regarding the default class loader so complex applications or environments (looking at you OSGi) may require some modifications but without complete error reports, it is not possible for me to know if Py4J or the user code must be changed.

johann-petrak:
(Just using github actions for unit testing or linting instead of Travis.com may be the easier and faster way forward, at least that was my experience after struggling with travis for some time)

Thrameos:
If it is any help JPype switched from travis to azure a while back.  We have pretty good patterns to build and test many architectures including 8 and 11.  See ``.azure``  in the JPype tree.  We also have building modules for many architeture release down.  But given I don't see a C portion this likely isnt needed here.

As for Java 9+, the module system was added which controls access to reflection by adding a new security model.  The solution may be as easy as adding a module-info.java to the base of the source tree and modifying the build pattern to enable modules.  Unfortunately I dont have this one down as JPype uses C JNI reflection and thus may have different access rules. 

[...]

PR #453, opened 2021-11-25
Add support for Java 17

torokati44:
This tries to address #403 and a few related issues, by implementing Option 3 from https://github.com/bartdag/py4j/issues/403#issuecomment-782728385.

Unfortunately(?) this raises the minimum JDK version to 9.
I agree that it's not possible to properly solve this, while also supporting both Java 8 and Java 17, with a single build.

I also gave Option 1 a try, but that way not even `getCallbackClient()` was accessible, so I didn't pursue it further.

### DIFF
--- a/py4j-java/build.gradle
+++ b/py4j-java/build.gradle
@@ -126,8 +126,8 @@ if (jdk6BootClasspath) {
 }
 
 tasks.withType(JavaCompile) {
-    sourceCompatibility = "1.6"
-    targetCompatibility = "1.6"
+    sourceCompatibility = "9"
+    targetCompatibility = "9"
     options.fork = true
     options.encoding = "UTF-8"
 }
--- a/py4j-java/gradle.properties
+++ b/py4j-java/gradle.properties
@@ -1,8 +1,8 @@
 group=py4j
 bundleVersion=0.10.9.3
 version=0.10.9.3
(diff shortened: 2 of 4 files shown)


tueda:
If it works only with JDK 9+, then should `sourceCompatibility` and `targetCompatibility` be changed to `9`? I find them in `py4j-java/gradle.properties` and `py4j-java/build.gradle`.

torokati44:
I suppose...

JoshRosen:
Apache Spark relies on Py4J and (as of now) has not announced plans to drop Java 8 support.

I'm wondering if we can retain Java 8 support by using either reflection or MethodHandles to use `trySetAccessible` only if it's available and continue running the existing code on Java 8. This idea is inspired by a similar proposal in `lz4-java` to use MethodHandles to use Java 9+ features if available without having to rely on multi-version JARs: https://github.com/lz4/lz4-java/issues/183

cozek:
Hi, any updates regarding this?

HyukjinKwon:
I will work on this soon - probably in the next quarter.

HyukjinKwon:
Hey guys, we plan to work on it in this quarter. I will leave it closed this as a duplicate of https://github.com/py4j/py4j/issues/485. Let's track it there.

PR #459, opened 2022-02-28
Adds venv and .idea to .gitignore

HyukjinKwon:
This PR proposes to add `venv` and `.idea` to `.gitignore`. 

`venv` can be created by [virtual environment](https://docs.python.org/3/library/venv.html)
`.idea` is created by [PyCharm](https://www.jetbrains.com/pycharm/promo/?source=google&medium=cpc&campaign=14123077093&gclid=Cj0KCQiA3-yQBhD3ARIsAHuHT64XonV3K7ORd-uilTj9b3YbqqapmULnnXumNYyHEayQe4erRRJuT7AaAt7xEALw_wcB) and [IntelliJ](https://www.jetbrains.com/idea/)

Both are metadata, and shouldn't be managed via Py4J source codes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .DS_Store
 /build
+venv
+.idea


HyukjinKwon:
Will take a look for the test failure shortly.

bartdag:
Hi @HyukjinKwon,

Considering that our .gitignore file already refers to .DS_Store and .swp (VIM), we can add .idea and venv.

I think it's generally better practice to put this in your global gitignore file (and we could thus remove .DS_Store and .swp)?

I really don't mind though, just wanted to know your experience with other projects.

HyukjinKwon:
I was thinking making the project env setup self-contained and portable .. but what you said makes sense too. I don't mind closing this either 👍 . I will leave it to you.

bartdag:
Considering that we have Eclipse users, we may end up adding eclipse, vs-code, and a bunch of other IDEs in the future so I would close for now (and I will probably open a PR to remove mac and vim files). Thanks for bringing this up!

PR #462, opened 2022-02-28
Deflake GatewayServerTest.testListener

HyukjinKwon:
This PR deflakes the `GatewayServerTest.testListener` test by simply using a different port. The test seems to be flaky because GitHub Actions machines are somewhat slow, and the next test gets triggered when the ports used in the previous tests are not closed properly yet.

We can leverage wait-condition but this PR just simply works around by using a different port since that doesn't affect `GatewayServerTest.testListener`'s test coverage.

See https://github.com/py4j/py4j/runs/5353919319?check_suite_focus=true

```
...
py4j.GatewayServerTest > testListener FAILED

    py4j.Py4JNetworkException at GatewayServerTest.java:82
130 tests completed, 1 failed
        Caused by: java.net.BindException at GatewayServerTest.java:82
...
```

### DIFF
--- a/py4j-java/src/test/java/py4j/GatewayServerTest.java
+++ b/py4j-java/src/test/java/py4j/GatewayServerTest.java
@@ -77,7 +77,8 @@ public class GatewayServerTest {
 	@Test
 	public void testListener() {
 		TestListener listener = new TestListener();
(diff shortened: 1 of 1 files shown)


bartdag (approved):
Thanks! Good to go!

PR #467, opened 2022-03-03
Keep CNAME in GitHub Pages

HyukjinKwon:
We should keep `CNAME`. Otherwise, it will be removed, see also https://github.com/py4j/py4j/commit/bdf851b14e45fdc4b216721bd070bf647bac8c9a


HyukjinKwon:
I just tested by repeating old and new workflows. Seems all good now. Just to extra clarify, there won't be any commit if there's no change in doc. I intentionally made some commits just now to test.

PR #463, opened 2022-02-28
Test and publish documentation via GitHub Actions and GitHub Pages

HyukjinKwon:
This PR proposes to add a GitHub workflow that:
- Tests (for PR) documentation
- Publishes (for commits in `master` branch) documentation

Current domain is https://py4j.github.io/py4j/. With https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/about-custom-domains-and-github-pages, https://py4j.org/ will be used shortly.

PR #468, opened 2022-03-03
Use automatically pickled in the ephemeral port range at ProxyTest

HyukjinKwon:
`testSayHello` at `ProxyTest` is very flaky:

```
py4j.ProxyTest > testSayHello FAILED
    py4j.Py4JNetworkException at ProxyTest.java:53
        Caused by: java.net.BindException at ProxyTest.java:53
```

The reason is that the previous port number in use is not properly closed yet, but this test attempts to use it.

This PR proposes to use automatically pickled in the ephemeral port range at `ProxyTest` by specifying `0`.

### DIFF
--- a/py4j-java/src/test/java/py4j/ProxyTest.java
+++ b/py4j-java/src/test/java/py4j/ProxyTest.java
@@ -48,15 +48,26 @@ public class ProxyTest {
 	public void setup() {
 		// GatewayServer.turnLoggingOn();
 		entry = new InterfaceEntry();
-		gServer = new GatewayServer(entry);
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
I will run tests some more time to see if this can fix the flakiness.

HyukjinKwon:
I ran 6 times. 5 times passed. one failed due to https://github.com/py4j/py4j/pull/463#issuecomment-1054080032 in which @itholic is investigating.

It should be ready for a look @bartdag

bartdag (approved):
Using a dynamic port should make these tests more robust indeed. Thanks!

PR #471, opened 2022-03-07
Clean up the leftover connection for finished threads in pinned thread mode

HyukjinKwon:
## What is the problem?

Correctly, there is resource leak when using the pinned thread mode (see also https://github.com/apache/spark/pull/24898).

For example, if you repeat the codes below multiple times to create Py4J connections in multiple threads,

```python
# PySpark application
import threading

def print_prop():
    # Py4J connection is used under the hood.
    print(spark.sparkContext.getLocalProperty("a"))

threading.Thread(target=print_prop).start()
```

the number of leftover connections grows:

```python
spark._jvm._gateway_client.deque
deque([<py4j.clientserver.ClientServerConnection object at 0x7fdc60170940>, <py4j.clientserver.ClientServerConnection object at 0x7fdca011e760>, <py4j.clientserver.ClientServerConnection object at 0x7fdcb01acdc0>, <py4j.clientserver.ClientServerConnection object at 0x7fdc60170100>, <py4j.clientserver.ClientServerConnection object at 0x7fdcb0232d30>])
```

In the environment where multiple threads are used without a pool, it easily causes "Too many files open" due to the lack of file descriptors (as they are all occupied by unclosed sockets).

## How do you fix?

This PR adds another variable to thread local that cleans up the connection right before the thread is finished. We need it as a separate thread local because `connection` is NOT cleaned because the reference is being held at `JavaClient.deque`.

See also https://github.com/py4j/py4j/commit/50fe45e46ca3c31f8ae140a3a12448cd5132ca46 for more details.

### DIFF
--- a/py4j-python/src/py4j/clientserver.py
+++ b/py4j-python/src/py4j/clientserver.py
@@ -9,7 +9,7 @@ code calls some Java code, the Java code will be executed in the UI thread.
 
 from __future__ import unicode_literals, absolute_import
 
-from collections import deque
+from collections import deque, Callable
 import logging
 import socket
 from threading import local, Thread
@@ -244,7 +244,10 @@ class JavaClient(GatewayClient):
         :param connection: The ClientServerConnection to associate with the
             current thread.
         """
-        self.thread_connection.connection = weakref.ref(connection)
+        conn = weakref.ref(connection)
+        self.thread_connection._cleaner = (
+            ThreadLocalConnectionFinalizer(conn, self.deque))
+        self.thread_connection.connection = conn
 
     def shutdown_gateway(self):
         try:
@@ -298,6 +301,39 @@ class JavaClient(GatewayClient):
 
     def _create_connection_guard(self, connection):
         return ClientServerConnectionGuard(self, connection)
+
+
+class ThreadLocalConnectionFinalizer(object):
+    """Cleans :class:`ClientServerConnection` held by a thread local by
+    closing it properly and removing it from the :class:`JavaClient`
+    deque. Right before the Python thread is terminated, this
+    instance will be garbage-collected, which triggers a call
+    to __del__  that contains the cleanup logic.
+    """
+    def __init__(self, connection, dequeue):
+        assert (
+            isinstance(connection, Callable) and
+            connection() is not None and
+            isinstance(connection(), ClientServerConnection))
+        self.connection = connection
+        self.deque = dequeue
+
+    def __del__(self):
+        """Removes the connection associated with the current thread
+        from the deque.
+
+        Expected to be called when the thread that started the
+        connection is garbage-collected.
+        """
+        conn = self.connection()
+        if conn is not None:
+            try:
+                # This dequeue is thread-safe, and shared across other
+                # threads.
+                self.deque.remove(conn)
+            except ValueError:
+                # Should never reach this point
+                pass
 
 
 class ClientServerConnectionGuard(GatewayConnectionGuard):
@@ -602,6 +638,12 @@ class ClientServerConnection(object):
             params.append(param)
             temp = smart_decode(input.readline())[:-1]
         return params
+
+    def __del__(self):
+        # In case new connection is set via
+        # `JavaClient.set_thread_connection`, this connection will be
+        # garbage-collected with closing the underlying socket properly.
+        self.close()
 
 
 class ClientServer(JavaGateway):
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
cc @bartdag I badly wanted to make this fix. I am happy that I am finally able to submit this PR.

py4j-python/src/py4j/tests/client_server_test.py
```diff
@@ -147,6 +147,44 @@ def testSendObjects(self):
             client_server.shutdown()
             self.assertEquals(1000, hello.calls)
 
+    def testCleanConnections(self):
```
HyukjinKwon, line 150: This passes in my local. I will take a look tmr.

WeichenXu123:
@HyukjinKwon awesome work to fix this!

HyukjinKwon:
I can't reproduce in local .. I will investigate it in CI.

py4j-python/src/py4j/clientserver.py
```diff
@@ -140,7 +140,7 @@ def __init__(
             ssl_context=None, auto_gc=False,
             accept_timeout=DEFAULT_ACCEPT_TIMEOUT_PLACEHOLDER,
             read_timeout=None, propagate_java_exceptions=False,
-            auth_token=None):
+            auth_token=None, auto_gc_connections=False):
```
HyukjinKwon, line 143: I checked other tests pass when `auto_gc_connections` is `True` except a couple of tests in `py4j.tests.memory_leak_test:ClientServerTest` which assumes several `ClientServerConnection`s are not garbage-collected after shutting down the client server.

HyukjinKwon:
This is WIP, and I am investigating more. Python 2 has a different behaviour on GC.

HyukjinKwon:
it should be ready for a look now.

HyukjinKwon:
it should be ready for a look now.

py4j-python/src/py4j/clientserver.py
```diff
@@ -203,6 +203,32 @@ class JavaClient(GatewayClient):
     ClientServerConnection per thread.
     """
 
+    class ThreadLocalConnectionCleaner(object):
```
bartdag, line 206: Should it be ThreadLocalConnectionFinalizer to better indicate its purpose (like the ThreadSafeFinalizer)?

py4j-python/src/py4j/clientserver.py
```diff
@@ -203,6 +203,32 @@ class JavaClient(GatewayClient):
...
+        """Cleans :class:`ClientServerConnection` in the thread local with
+        closing it properly, and removes it from the deque at
```
bartdag, line 208: Suggested change (but English is my second language too so take it with a grain of salt):

```
Cleans ... held by a thread local by closing it properly and removing it from the :class:`JavaClient` deque.

Right before the Python thread is terminated, this instance will be garbage-collected, which triggers a call to __del__  that contains the cleanup logic.
```

py4j-python/src/py4j/clientserver.py
```diff
@@ -203,6 +203,32 @@ class JavaClient(GatewayClient):
...
+        def __init__(self, connection, dequeue):
+            assert (
+                isinstance(connection, Callable) and
+                connection() is not None and
+                isinstance(connection(), ClientServerConnection))
+            self.connection = connection
+            self.deque = dequeue
+
+        def __del__(self):
+            conn = self.connection()
```
bartdag, line 222: I would add a docstring.

```
"""Removes the connection associated with the current thread from the deque.

Expected to be called when the thread that started the connection is garbage-collected.
"""
```

The second sentence was what I struggled to understand while reviewing the code.

py4j-python/src/py4j/clientserver.py
bartdag: Also, does this class need to be inside JavaClient? This is a minor comment, but I try to only create internal structure (internal function and internal class) if they need to refer to the outer context.

bartdag (commented):
Awesome work. It took me a while to understand what was causing the issue in the first place.

I made a few suggestions, but I do not feel strongly towards them. Feel free to take what you want from them!

Thinking about the root cause, in the future, we may want to add a method to allow threads to explicitly tell when their work is done. With the default gateway approach, you can always call GatewayClient.close() to close all connections, but this does not make sense in the pinned thread model (you want to close your connection, not all connections).

That being said, your solution is the only sane way to stay backward compatible, so kudos!!

py4j-python/src/py4j/clientserver.py
HyukjinKwon: Let me take a closer look ..

HyukjinKwon:
Merged!

Issue #473, opened 2022-03-16
🐛 DeprecationWarning introduced in commit 799ac4e


gabe-l-hart:
## Description

The changes made in [799ac4e](https://github.com/py4j/py4j/commit/799ac4eca035a7e90accae06d006a1fe4ec3203c) added [this import](https://github.com/py4j/py4j/blame/master/py4j-python/src/py4j/clientserver.py#L12). In turn, this triggers the following `DeprecationWarning` from `collections`:

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
```

The core python `collections` library has [fully removed this functionality](https://github.com/python/cpython/pull/14171/files).

## Proposed Fix

I believe that the correct solution is to change the imports from

```py
from collections import deque, Callable
```

to

```py
from collections import deque
from collections.abc import Callable
```

gabe-l-hart:
I can confirm that with `0.10.9.5` we're not seeing the `DeprecationWarning` anymore on `3.8.8` and nothing seems to have broken. Thanks!

PR #474, opened 2022-03-16
Import Callable from collections.abc

gabe-l-hart:
## Description

This PR attempts to address https://github.com/py4j/py4j/issues/473

### DIFF
--- a/py4j-python/src/py4j/clientserver.py
+++ b/py4j-python/src/py4j/clientserver.py
@@ -10,7 +10,6 @@ code calls some Java code, the Java code will be executed in the UI thread.
 from __future__ import unicode_literals, absolute_import
 
 from collections import deque
-from collections.abc import Callable
 import logging
 import socket
 from threading import local, Thread
@@ -313,7 +312,7 @@ class ThreadLocalConnectionFinalizer(object):
     """
     def __init__(self, connection, dequeue):
         assert (
-            isinstance(connection, Callable) and
+            callable(connection) and
             connection() is not None and
             isinstance(connection(), ClientServerConnection))
         self.connection = connection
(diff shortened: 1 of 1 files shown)


py4j-python/src/py4j/clientserver.py
```diff
@@ -9,7 +9,8 @@
 
 from __future__ import unicode_literals, absolute_import
 
-from collections import deque, Callable
+from collections import deque
+from collections.abc import Callable
```
HyukjinKwon, line 13: The problem was that it breaks Python 2 so I used the deprecated one. Probably we can just remove the usage of `Callable`, and check if it is a weakref reference.

gabe-l-hart: That makes perfect sense! Sorry I missed that. I'll to get a python 2 compatible version up tomorrow.

HyukjinKwon: Nah, thanks for addressing this issue man.

gabe-l-hart: According to [the internet](https://stackoverflow.com/a/624939), it looks like the builtin `callable` should be a solution that works for all 2.X and >=3.2. Is that a safe range for `py4j` at this point?

gabe-l-hart: I've verified locally on `2.7`, `3.9`, and `3.10`. I've also verified that in `3.10` an `ImportError` is raised on `from collections import Callable`

HyukjinKwon: Yeah it broke python 3.10 support. Or we can even just remove the assert for now.

HyukjinKwon:
Actually https://github.com/py4j/py4j/pull/475 suggested the fix .. let me merge that one and close this one - will add you as a co-author of https://github.com/py4j/py4j/pull/475.

Thanks for investigation @gabe-l-hart!

PR #475, opened 2022-03-17
Fix import error on Python 3.10+

shadchin:


HyukjinKwon (commented):
Thanks man. Can we actually just avoid using `Callable`? It's only for `assert`. We can just check if it's a weakref or not in the assert.

HyukjinKwon:
We should also set up it w/ Python 3.10 too .. I didn't realise that it's removed in Python 3.10.

shadchin:
Maybe switch on `callable()` ? https://docs.python.org/3/library/functions.html#callable

HyukjinKwon (approved):

HyukjinKwon:
I am adding @gabe-l-hart as a co-author since he investigated/reported this issue first, and made an almost similar fix w/ this one at https://github.com/py4j/py4j/pull/474.

PR #490, opened 2022-07-08
Add JDK11 support to CI.

ueshin:
Adds JDK11 support to CI.

The build should still be done with JDK 8 and test with JDK 8 or 11.

- Add 11 to `java-version` metrix to use for PyTest
- Disables `findbugs` plugin that doesn't support JDK>=9
- Upgrade `jacoco` plugin to 0.8.7


HyukjinKwon:
Let me close this since we're now testing JDK 11/17

PR #491, opened 2022-07-13
Add support for Java 17.

### LINKED ISSUE #485: JDK 11 and 17 support
See
- https://github.com/py4j/py4j/pull/453
- https://github.com/py4j/py4j/issues/403
- https://github.com/py4j/py4j/issues/448

We should:
- Set up the CI for JDK 11/17
- See how we'd like to release. e.g.) compile with JDK 8 to make one release working with both 11/17 (like Apache Spark does as an example).

ueshin:
Based on #453, introduces `ReflectionShim` for older versions of JDK that don't have `AccessibleObject.trySetAccessible` to fallback to `setAccessible`.

Resolves #485.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,16 +14,16 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-18.04, windows-2019 ]
-        java-version: [ 8 ]
+        java-version: [ 8, 11, 17 ]
         python-version: [ '2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10' ]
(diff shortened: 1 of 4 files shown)


HyukjinKwon (approved):
LGTM if tests pass

ueshin:
> LGTM if tests pass

Tests can pass if we run multiple times... The tests on windows seem so flaky.

HyukjinKwon:
Yeah, it's flaky. We should actually fix them. I don't think this PR causes the flakiness.

ueshin:
Tests passed after several re-runs... 😓

PR #492, opened 2022-08-08
Detach Java objects only when _detach method exists

HyukjinKwon:
This is more a bug or safeguard.

After conversion of arguments (via our `Converter.convert` interface in Py4J), the returned argument might not be a plain `JavaObject`. For example, `JavaObject(java.lang.Integer)` would be converted to `int` automatically, see also https://github.com/py4j/py4j/issues/163.

However, the current codebase requires it to be a `JavaObject` by assuming `_detach` method exists (to garbage collect the instance). In fact, calling a Java method with these Python primitives are valid in Py4J, so it makes sense to allow passing returning primitive types via `Converter.convert`.

Therefore, this PR proposes to call `_detach` only when it exists, and delegates the type checking into actual method invocation that is consistent with calling the usual JVM methods via Py4J.

I manually tested, and will add the integration test into PySpark side. It's a bit tricky to add a unittest, and it will be tested together with PySpark.

See also https://github.com/apache/spark/pull/32955#discussion_r658315864

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -1305,7 +1305,8 @@ class JavaMember(object):
             answer, self.gateway_client, self.target_id, self.name)
 
         for temp_arg in temp_args:
-            temp_arg._detach()
+            if hasattr(temp_arg, "_detach"):
+                temp_arg._detach()
 
         return connection
 
@@ -1322,7 +1323,8 @@ class JavaMember(object):
             answer, self.gateway_client, self.target_id, self.name)
 
         for temp_arg in temp_args:
-            temp_arg._detach()
+            if hasattr(temp_arg, "_detach"):
+                temp_arg._detach()
 
         return return_value
 
@@ -1586,7 +1588,8 @@ class JavaClass(object):
             answer, self._gateway_client, None, self._fqn)
 
         for temp_arg in temp_args:
-            temp_arg._detach()
+            if hasattr(temp_arg, "_detach"):
+                temp_arg._detach()
 
         return return_value
 


HyukjinKwon:
cc @ueshin @bartdag @xinrong-meng FYI

py4j-python/src/py4j/java_gateway.py
```diff
@@ -1305,7 +1305,8 @@ def stream(self, *args):
             answer, self.gateway_client, self.target_id, self.name)
 
         for temp_arg in temp_args:
-            temp_arg._detach()
+            if hasattr(temp_arg, "_detach"):
```
ueshin, line 1308: This and the below should be `isinstance(temp_arg, JavaObject)`?

ueshin (commented):
I'm also wondering https://github.com/py4j/py4j/blob/529b6af16db4fdef265514ef57978650c2475dc2/py4j-python/src/py4j/java_gateway.py#L2067-L2078 will ever be called with primitive values.

We might also want to add the if clause here.

py4j-python/src/py4j/java_gateway.py
xinrong-meng: Will that `JavaObject` assumption break downstream codes that have customized objects with `_detach` defined?
Seems rare though.

ueshin: I think that's rather a bug if we call `_detach` on arbitrary objects unrelated to `py4j`.

xinrong-meng: Sorry I meant external developers' codes that extend `py4j`, which customized a class with `_detach` method.

HyukjinKwon: I think either way is fine. In Py4J both `isinstance` and `hasattr` usages are mixed anyway. Some interfaces like Converter just assumes that it has `convert` method too.

HyukjinKwon:
re: https://github.com/py4j/py4j/pull/492#pullrequestreview-1065577974

Maybe yes. But I think it's pretty unlikely because that's explicit `detach` which can be selectively applied whereas here is implicit `_detach` that is always applied.

HyukjinKwon:
Let me get this merged first. let me know if you think differerntly.

PR #498, opened 2022-08-24
Give py4j thread an explicit name, mark them as daemon

alexarchambault:
This PR makes py4j give more explicit names to the threads it creates (like `py4j-…` rather than the default `Thread-…`), which makes it easier to inspect JVM processes where py4j is running.

It also [marks those threads as daemon](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setDaemon-boolean-), so that exiting the main thread of the JVM, by reaching the end of the `main` method, actually makes the JVM exit (provided no other non-daemon threads were created by users).

### DIFF
--- a/py4j-java/src/main/java/py4j/ThreadUtil.java
+++ b/py4j-java/src/main/java/py4j/ThreadUtil.java
@@ -1,0 +1,31 @@
+package py4j;
+
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * <p>
+ * Thread utility class providing a method to create threads.
+ * </p>
+ *
+ * @author Alex Archambault
+ *
+ */
+public class ThreadUtil {
+
+    private static ThreadFactory factory = new ThreadFactory() {
+        private final AtomicInteger counter = new AtomicInteger();
+        @Override
+        public Thread newThread(Runnable r) {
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
oh yeah, probably it's a good idea to name them

py4j-java/src/main/java/py4j/ThreadUtil.java
```diff
@@ -0,0 +1,31 @@
+package py4j;
```
HyukjinKwon, line 1: I think you don't have to bother with creating a util here. Let's just fix them inplace at `GatewayServer`, `ClientServerConnection`, and let's name them.

py4j-java/src/main/java/py4j/ThreadUtil.java
```diff
@@ -0,0 +1,31 @@
...
+public class ThreadUtil {
+
+    private static ThreadFactory factory = new ThreadFactory() {
+        private final AtomicInteger counter = new AtomicInteger();
+        @Override
+        public Thread newThread(Runnable r) {
+            int index = counter.incrementAndGet();
+            String name = "py4j-" + index;
+            Thread thread = new Thread(r, name);
+            thread.setDaemon(true);
```
HyukjinKwon, line 23: but why should we make it a daemon though?

PR #499, opened 2022-09-07
Root logger used in send command

Baunsgaard:
This PR fixes the logger used in exception handling when sending commands.
Previously the root logger was used, and with this fix the py4j specific logger is used.


HyukjinKwon (approved):
Looks good from a cursory look. cc @ueshin FYI

Baunsgaard:
@HyukjinKwon 
I ran into this issue again in my code base.
and now the new merge conflict is resolved as well.

Baunsgaard:
@HyukjinKwon  ? I ran into this the other day again, any plans on merging it?

Baunsgaard:
Maybe there is an update now after years @Tagar ?

Tagar:
Good catch! I see @HyukjinKwon already approved. Thanks for the fix @Baunsgaard

Issue #374, opened 2019-07-18
Long type Map key problem with Python 3


marianp123:
I have this problem when running with Python 3.7, it works fine with 2.7.
I make a call from Python to a Java method returning a java.util.Map with Long type key. JavaMap is returned with entries having correct keys (as int-s as there is no long type in Python 3), but None values. 
I understand that Py4j converts Python int values to Java long for values over the java integer range, which is reasonable, but this case is different and should be handled by Py4J in my opinion.

`public Map javaMethod() {
  Map map = new HashMap();
  map.put(25L, 26);
  return map;
}

_dict = gateway.JavaClass.javaMethod()
print(_dict)  # prints: {25: None}

...
  map.put(((long) Integer.MAX_VALUE), 26);

_dict = gateway.JavaClass.javaMethod()
print(_dict)  # prints: {2147483647: None}

...
  map.put(((long) Integer.MAX_VALUE) +1, 26);

_dict = gateway.JavaClass.javaMethod()
print(_dict)  # prints: {2147483648: 26}

`

qvad:
@marianp123 I don't think there is any chance to catch this in the py4j side. In case of generic types, you have no information about value type in runtime. Otherwise, while you calling Java method from py4j you have all information about method arg types.

So the solution with looping through all number types while getting generic collection value looks strange. (You'll have the same problem with Byte, Short etc...)

I fixed this behaviour by creating special method on Java side and this is very ugly

    @SuppressWarnings("unchecked")
    public Object getTypedKey(Cache cache, Object key, String type) throws ClassNotFoundException {
        switch (type.toLowerCase()) {
            case "long":
                return cache.get(createLong((Number)key));

            case "byte":
                return cache.get(createByte((Number)key));

            case "short":
                return cache.get(createShort((Number)key));

            default:
                return cache.get(Class.forName(type).cast(key));

        }
    }

marianp123:
@qvad Java reflection provides information about a collection's generic types. Py4j returns a JavaMap->Dictionary object. Since it's doing the conversion, it should take care of the special case. The JavaMap object contains the key, but no value. I have no option to retrieve the value unless I change the Java side method. I had this conversion problem in other cases, I am not sure if long wouldn't be a better default than integer. it's a tricky problem. an option to customize the behavior per call might be the best.

bartdag:
Hi, this is currently a limitation of Py4J and there is no easy way to get around that.

[...]

PR #500, opened 2022-09-28
Add TypeInt for Specifying Integer Variant from Python to Java

srilman:
This PR addresses issue #374. It provides a way for users to specify how they would like Python integers to be converted into Java (whether to treat them like an int, short, or long). This PR currently only adds an integer and long variant, since Py4J does not have general support for shorts.

### DIFF
--- a/py4j-python/src/py4j/protocol.py
+++ b/py4j-python/src/py4j/protocol.py
@@ -21,6 +21,8 @@ from __future__ import unicode_literals, absolute_import
 from base64 import standard_b64encode, standard_b64decode
 
 from decimal import Decimal
+from enum import Enum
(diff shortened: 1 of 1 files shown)


HyukjinKwon (commented):
Can we add a test? Also, does this work with other type hints?

srilman:
Do you have some recommendations for tests? I was thinking of adding a a test to `java_array_test.py` of creating arrays using TypeInt, and maybe some basic examples of communicating from Python to Java. Are there any good examples of tests that do that?

Also, what do you mean by "other type hints"?

njriasan:
@HyukjinKwon What needs to be done with this PR to get it merged?

Issue #516, opened 2023-04-25
Performance Issue due to TCP Nagle's Algorithm write,write,read Situation


trohwer:
I am aware, that passing byte arrays via py4j is not (supposed to be) very efficient (although some copy operations along the path could be avoided).
Nevertheless I was initially surprised by the following performance difference (here in py4j-0.10.9.5.jar included in PySpark on Linux):

import time

b=spark.sparkContext._jvm.java.nio.ByteBuffer.allocate(4096)
t0=time.time()
for i in range(0,100):
  u=b.array()
print(time.time()-t0)

# 0.04267597198486328

b=spark.sparkContext._jvm.java.nio.ByteBuffer.allocate(8192)
t0=time.time()
for i in range(0,100):
  u=b.array()
print(time.time()-t0)

# 4.404087543487549

It turns out that the code suffers from Nagle algorithm here. E.g. in the CallCommand

		writer.write(returnCommand);
		writer.flush();
		
if writing returnCommand exceeds the buffer of the BufferedWriter, there are two writes to the socket output.

After disabling the Nagle algorithm for loopback sockets by adding the following in ClientServerConnection.java

    super();
    this.socket = socket;

    // added
    if (socket.getLocalAddress().isLoopbackAddress()) socket.setTcpNoDelay(true);

    this.reader = new BufferedReader(new InputStreamReader(socket.getInputStream(), Charset.forName("UTF-8")));

I get the following run time measurements:

0.047772884368896484
0.07696914672851562

I think, that for loopback sockets disabling the algorithm does not have any disadvantages, since buffering occurs in the BufferedWriter. Possibly one could disable it in general.

PR #517, opened 2023-04-25
Disable Nagle's algorithm on loopback sockets.

trohwer:
See issue 516.

### DIFF
--- a/py4j-java/src/main/java/py4j/ClientServerConnection.java
+++ b/py4j-java/src/main/java/py4j/ClientServerConnection.java
@@ -72,6 +72,10 @@ public class ClientServerConnection implements Py4JServerConnection, Py4JClientC
 					throws IOException {
 		super();
 		this.socket = socket;
+
+		// disable Nagle's algorithm on loopback connection to avoid performance issues due to write,write,read situations
+		if (socket.getLocalAddress().isLoopbackAddress()) socket.setTcpNoDelay(true);
+
 		this.reader = new BufferedReader(new InputStreamReader(socket.getInputStream(), Charset.forName("UTF-8")));
 		this.writer = new BufferedWriter(new OutputStreamWriter(socket.getOutputStream(), Charset.forName("UTF-8")));
 		this.commands = new HashMap<String, Command>();


HyukjinKwon:
Sorry for the delay. Now the tests are fixed per https://github.com/py4j/py4j/pull/519. Would you mind rebasing this one please?

trohwer:
Hello, I merged the main branch. Is it ok now?

py4j-java/src/main/java/py4j/ClientServerConnection.java
```diff
@@ -74,6 +74,10 @@ public ClientServerConnection(Gateway gateway, Socket socket, List<Class<? exten
 					throws IOException {
 		super();
 		this.socket = socket;
+
+		// disable Nagle's algorithm on loopback connection to avoid performance issues due to write,write,read situations
+		if (socket.getLocalAddress().isLoopbackAddress()) socket.setTcpNoDelay(true);
```
HyukjinKwon, line 79: Do we have a rough evidence for this performance improvement? e.g., benchmark.

HyukjinKwon: And what's the impact of disabling this?

trohwer: See https://github.com/py4j/py4j/issues/516 for background and a measurement for a slow case. For the loopback interface I would assume, there are no negative performance impacts for disabling Nagle. I would have assumed it to be the default on loopback, but that seems not to be always the case.

Tagar:
Picked this up to validate the impact using the perf framework being developed in #578 — wanted to put concrete numbers on what the change does on a realistic py4j workload. Running on macOS arm64 (Apple M4 Pro, loopback only), 50 measured rounds per scenario, full bootstrap + Mann-Whitney + KS test.

Six macro scenarios spanning single-threaded, multi-threaded (1/4/16/50 threads), bulk-list ops, and Java→Python callbacks:

| Scenario | Concurrency | Δ median | Δ p99.9 | Tail ratio | KS p | Verdict |
|---|---|---|---|---|---|---|
| X1-1 — concurrent_1_thread | 1 thread | −0.3% | −0.3% | 1.00× | 6.7e-4 | neutral |
| X1-16 — concurrent_16_threads | 16 threads | −0.5% | **+9.0%** | 1.09× | 0.112 | inconclusive (possible tail regression) |
| X2-10k — iterate_javalist_10k | 1 thread | +1.0% | −1.6% | 0.99× | 3.0e-4 | neutral |
| X3-1k — listconverter_1k | 1 thread | +0.6% | +0.7% | 1.01× | 0.003 | neutral |
| X4 — callback_sort_100_items | Java→Python callbacks | +0.3% | +2.3% | 1.01× | 0.178 | inconclusive |
| X6 — pool_saturation_50_threads | 50 threads | −0.6% | −4.5% | 0.96× | 0.112 | inconclusive |

"Verdict" is the framework's combined criterion: confident faster/regression requires Mann-Whitney p < 0.01, bootstrap 95% CI excluding zero, AND |Hodges-Lehmann delta| ≥ 5%.

A few observations, not conclusions:

1. **No scenario clears the 5% practical-significance bar.** KS detects distribution-level differences on X1-1 / X2-10k / X3-1k (p ≪ 0.01), so the change *is* measurable — it just isn't a 5%+ effect on these particular workloads.
2. **X1-16 shows a possible tail regression**: p99 / p99.9 both grow ~9%, tail ratio 1.09×. KS p = 0.112 doesn't clear significance with N=50, but the direction is consistent and worth re-running with more samples. Plausible mechanism: 16 concurrent connections sending small request/response messages may benefit from Nagle's coalescing — disabling it produces more under-MSS packets and more kernel work per call.
3. **My measurement scope is narrow.** macOS loopback, Apple Silicon, request/response macro scenarios. The Nagle / delayed-ACK interaction can look quite different on Linux loopback, and a streaming or long-poll workload may show the benefit clearly even when these don't. Very possible the workload you originally observed the improvement on falls into a regime these scenarios don't model.

To reproduce or cross-check on your end:

```bash
git clone https://github.com/Tagar/py4j.git
cd py4j
git checkout perf-framework-v2

# Baseline on master
python -m py4j.tests.perf \
  --only X1-1,X1-16,X2-10k,X3-1k,X4,X6 \
  --save baseline.json --save-compact

# Cherry-pick this PR's commit (or check it out) and re-run
git cherry-pick <this-PR-commit>
python -m py4j.tests.perf \
  --only X1-1,X1-16,X2-10k,X3-1k,X4,X6 \
  --compare baseline.json
```

If you do run it and see a different pattern — especially on Linux or in a specific workload — happy to dig into the divergence. And if there's a py4j usage pattern that should benefit but isn't represented in the macro scenarios above, I'd be glad to add it. Not arguing against the change; just wanted to share what the framework measures so the discussion has concrete numbers either way.

PR #538, opened 2024-03-18
Improve Py4j command cancellation by shutting down the JVM side of the socket

sadikovi:
This PR improves PySpark command cancellation by shutting down the JVM side of the connection:
- JVM thread is shut down.
- All of the Py4j connections that are related to the local or remote port are also shut down.

The idea is to find all of the open sockets in JVM on KeyboardInterrupt and then send the special "CancelCommand" with local and remote ports to cancel the matching socket in JVM. This helps to cancel commands in a notebook correctly in cases when the Java side executes code asynchronously. In those cases, I found KeyboardInterrupt only terminates the Python side of the connection.


sadikovi:
cc @HyukjinKwon

HyukjinKwon (approved):

sadikovi:
@HyukjinKwon Can you review again? I updated the tests to not use Mockito.

PR #542, opened 2024-04-06
Add method to terminate JVM process

romik2512:
I use Payspark in my work, which uses py4j to connect PVM and JVM processes. 
I often need to restart the Spark JVM driver process and redefine the amount of memory for it. The existing methods do not ensure the completion of the JVM process. The only way to achieve this is to complete the Python process, which we cannot afford.

The current PR adds a new method to the JavaGateway class, which will allow JVM processes to terminate. Its logic is to call System.exit() on the JVM side without waiting for a return response.

These changes are based on an article (in Russian) where the current problem was fixed manually.
Article: https://habr.com/ru/companies/sberbank/articles/805285


dolfinus:
Duplicate: #541

PR #543, opened 2024-04-15
[TESTS-ONLY] Fix Flaky `GatewayServerTest`

ahirreddy:
- This PR fixes the GatewayServerTest to handle a race condition in the GatewayServer's handling of connections with invalid secrets.
- The server may either return an appropriate error message (this is what the test expected pre-patch) or the server may return no response (which becomes a `null` response)
- The test-case is updated to accept both responses as valid


HyukjinKwon:
Merged, thx.

PR #554, opened 2025-02-26
Iterator.next should not ignore other exceptions

HyukjinKwon:
It assumes `Py4JError` should always come from the iterator termination signal `NoSuchElementException`, which is not correct. We should not assume it means the iterator is terminated


ueshin (approved):
LGTM. Thanks for the fix!

PR #510, opened 2023-02-24
Issue 374 Add type hint

mucharafal:
Base on https://github.com/py4j/py4j/pull/500, if https://github.com/Bodo-inc/py4j/pull/1 is merged, then this may be closed.

PR #569, opened 2025-07-08
Fixes #374 -- Explicit TypeHint support

### LINKED ISSUE #374: Long type Map key problem with Python 3

praneeth527:
Fixes #374 
Added support for explicit Type conversion

Test cases are passed ✅ 
Screenshot 2025-07-08 at 11 37 49 AM

Existing/Reference PR -- #510

### DIFF
--- a/py4j-python/src/py4j/protocol.py
+++ b/py4j-python/src/py4j/protocol.py
@@ -190,7 +190,7 @@ class TypeHint:
         >>> gateway.jvm.SomeClass().methodLong(TypeHint(LONG_TYPE, 123))
 
     Args:
-        java_type: The target Java type (e.g., LONG_TYPE, INT_TYPE).
+        java_type: The target Java type (e.g., LONG_TYPE, INTEGER_TYPE).
         value: The Python value to be wrapped with the specified Java type.
     """
 


praneeth527:
Hi @HyukjinKwon,
Following up, Can you please take a look and let me know if any further changes required. Thanks

PR #575, opened 2026-03-15
perf: Remove Python 2 compat and optimize old calls to smart_decode

markjm:
Thanks for this project! I have a few perf changes that I think might be valuable to contribute back, so here is one!

Remove Python 2 compatibility layer since Python 2 is no longer supported. This simplifies the codebase, but more importantly **improves performance by eliminating unnecessary type checks and function call overhead**.

Since this project only supports py3.8 and above (https://github.com/py4j/py4j/blob/master/setup.py#L57), this is not a breaking change.

Changes:
- Mostly remove compat module usage from all main source files
- Replace smart_decode() with direct str() or .decode("utf-8") calls
- Use Python 3 built-ins directly (str, bytes, int, range, Queue, Thread)

Benchmark results (isolated microbenchmarks):
- str() vs smart_decode() for integers: 1.90x faster
- str() vs smart_decode() for booleans: 2.61x faster
- str() vs smart_decode() for Decimal: 1.75x faster
- .decode("utf-8") vs smart_decode() for stream reading: 2.01x faster
(micro-benchmarks available here https://gist.github.com/markjm/57cd1731abc0c1ce9820c413b90f992c)

These speedups are probably not game-changing, but given they are there for (now that py2 is totally dead) no reason at all, i figure we may as well take a free win however small.

The compat module is kept for backwards compatibility with any external code that may import from it, but is no longer used internally.

### DIFF
--- a/py4j-python/src/py4j/protocol.py
+++ b/py4j-python/src/py4j/protocol.py
@@ -16,16 +16,9 @@ Created on Oct 14, 2010
 
 :author: Barthelemy Dagenais
 """
-from __future__ import unicode_literals, absolute_import
-
 from base64 import standard_b64encode, standard_b64decode
 
 from decimal import Decimal
-
-from py4j.compat import (
-    long, basestring, unicode, bytearray2,
-    bytestr, isbytestr, isbytearray, ispython3bytestr,
-    bytetoint, bytetostr, strtobyte)
 
 
 JAVA_MAX_INT = 2147483647
@@ -159,7 +152,7 @@ DIR_JVMVIEW_SUBCOMMAND_NAME = "v\n"
 OUTPUT_CONVERTER = {
     NULL_TYPE: (lambda x, y: None),
     BOOLEAN_TYPE: (lambda value, y: value.lower() == "true"),
-    LONG_TYPE: (lambda value, y: long(value)),
+    LONG_TYPE: (lambda value, y: int(value)),
     DECIMAL_TYPE: (lambda value, y: Decimal(value)),
     INTEGER_TYPE: (lambda value, y: int(value)),
     BYTES_TYPE: (lambda value, y: decode_bytearray(value)),
@@ -185,7 +178,7 @@ def escape_new_line(original):
     :rtype: an escaped string
     """
     if original:
-        return smart_decode(original).replace("\\", "\\\\").\
+        return original.replace("\\", "\\\\").\
             replace("\r", "\\r").replace("\n", "\\n")
     else:
         return original
@@ -213,17 +206,16 @@ def unescape_new_line(escaped):
 
 
 def smart_decode(s):
-    if isinstance(s, unicode):
+    if isinstance(s, str):
         return s
-    elif isinstance(s, bytestr):
-        # Should never reach this case in Python 3
-        return unicode(s, "utf-8")
+    elif isinstance(s, bytes):
+        return s.decode("utf-8")
     else:
-        return unicode(s)
+        return str(s)
 
 
 def encode_float(float_value):
-    float_str = smart_decode(repr(float_value))
+    float_str = str(float_value)
     if float_str == "-inf":
         float_str = JAVA_NEGATIVE_INFINITY
     elif float_str == "inf":
@@ -234,16 +226,14 @@ def encode_float(float_value):
 
 
 def encode_bytearray(barray):
-    if isbytestr(barray):
-        return bytetostr(standard_b64encode(barray))
+    if isinstance(barray, bytes):
+        return standard_b64encode(barray).decode("ascii")
     else:
-        newbytestr = bytestr(barray)
-        return bytetostr(standard_b64encode(newbytestr))
+        return standard_b64encode(bytes(barray)).decode("ascii")
 
 
 def decode_bytearray(encoded):
-    new_bytes = strtobyte(encoded)
-    return bytearray2([bytetoint(b) for b in standard_b64decode(new_bytes)])
+    return bytearray(standard_b64decode(encoded.encode("ascii")))
 
 
 def is_python_proxy(parameter):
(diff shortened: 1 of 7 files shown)


markjm:
hi again @HyukjinKwon - let me know if there is anything else youd like to see here!

Tagar:
@markjm — took a look at this with the perf framework being developed in #578, wanted to put end-to-end numbers next to the isolated microbenchmarks in your description. Running on macOS arm64 (Apple M4 Pro, loopback only), idle machine. Framework's micro path (pytest-benchmark-driven) auto-scales rounds and iterations to keep timer resolution out of the noise; full bootstrap + Mann-Whitney + Kolmogorov-Smirnov stack on the per-round samples.

| Scenario | n (B/C) | Δ median | Δ p95 | Δ p99 | Δ p99.9 | Verdict |
|---|---|---|---|---|---|---|
| M1 — m1_static_call_no_args | 50/50 | +0.7% | +5.1% | +11.2% | +13.2% | inconclusive |
| M5a — m5a_encode_int | 428,578 / 626,610 | **−22.8%** | −24.2% | −23.5% | −11.6% | **faster** |
| M5b — m5b_encode_string | 1,951,194 / 1,951,194 | **−25.2%** | −25.8% | −26.8% | −26.5% | **faster** |

For M5a / M5b: bootstrap 95% CI is tight ([−23.0%, −22.7%] and [−25.7%, −24.6%]), Mann-Whitney p < 1e-200, P(better) = 1.00, KS p < 1e-200, tail ratio 0.73–0.76× — distribution moves cleanly toward faster, no asymmetric tail growth.

Observations:

1. **M5a / M5b — confident faster, no surprises.** The optimization gives a real ~22–25% median speedup on encode-int and encode-string microbenchmarks, with the whole distribution shifted, not just the median.
2. **Reconciling with the PR description's numbers.** Description shows ~1.9–2.6× on the isolated encoding path — that's consistent with the framework's ~1.30× end-to-end on M5a / M5b (220 → 170 ns and 384 → 287 ns medians). Encoding is one component of the per-call cost in the framework's microbench, so a 2× win on encoding alone surfaces as a ~25% win on the bundled scenario. No contradiction — your description shows the isolated change, the framework shows what shipped users would actually see.
3. **M1 — inconclusive, expected.** `m1_static_call_no_args` doesn't exercise the encoding hot path, so the verdict of "no confident change" is the correct outcome. P(same) = 0.99 confirms the change is approximately neutral on that path rather than a regression hiding under noise.

To reproduce or cross-check on your end:

```bash
git clone https://github.com/Tagar/py4j.git
cd py4j
git checkout perf-framework-v2

# Baseline on master
python -m py4j.tests.perf --only M1,M5a,M5b \
  --save baseline.json --save-compact

# Cherry-pick this PR's commit and re-run
git cherry-pick <this-PR-commit>
python -m py4j.tests.perf --only M1,M5a,M5b \
  --compare baseline.json
```

Two-line summary: the framework reproduces the optimization's benefit on the encoding path with no tail regressions or surprises. Looks like a clean net positive on macOS arm64; would be useful to also see on Linux before merge.

markjm:
seems about aligned with what you found, which i think makes sense just based on the kind of change this is:


# py4j perf comparison

**Baseline:** master (rev 1187a1c, 2026-05-14T04:50:43+00:00)
**Current:**  master (rev d2a8cdf, 2026-05-14T04:51:37+00:00)

| ID | Scenario | n (B/C) | Baseline median | Current median | Delta median | 95% CI | M-W p | P(better) | P(same) | P(worse) | Verdict |
|----|----------|---------|-----------------|----------------|--------------|--------|-------|-----------|---------|----------|---------|
| M1 | m1_static_call_no_args | 50/50 | 63.791 µs | 64.133 µs | +0.5% | [-5.0%, +5.0%] | 0.956 | 0.02 | 0.95 | 0.02 | inconclusive |
| M5a | m5a_encode_int | 1949334/1912019 | 427.496 ns | 348.066 ns | -18.6% | [-18.6%, -18.3%] | <1e-200 | 1.00 | 0.00 | 0.00 | **faster** |
| M5b | m5b_encode_string | 1058213/1736112 | 659.004 ns | 555.022 ns | -15.8% | [-15.8%, -15.5%] | <1e-200 | 1.00 | 0.00 | 0.00 | **faster** |

*Verdict: confident faster/regression requires Mann-Whitney p < 0.01, bootstrap 95% CI excluding zero, AND |Hodges-Lehmann delta| >= 5%. Otherwise: inconclusive.*
*P(better), P(same), P(worse) are region-of-practical-equivalence probabilities derived from the bootstrap distribution: fractions of resamples with Δ < -5%, |Δ| ≤ 5%, and Δ > +5% respectively. Human-meaningful direct probabilities — useful when the verdict is "inconclusive" but you want to know which side the evidence leans.*

### Per-percentile deltas

*Side-by-side relative changes at each percentile. A scenario whose median improves but whose p99 regresses signals a tail-latency regression — easy to miss if only the median is compared.*

| ID | Scenario | Δ median | Δ p95 | Δ p99 | Δ p99.9 |
|----|----------|----------|-------|-------|---------|
| M1 | m1_static_call_no_args | +0.5% | +11.3% | +14.1% | +16.5% |
| M5a | m5a_encode_int | -18.6% | -7.0% | -13.3% | -13.3% |
| M5b | m5b_encode_string | -15.8% | -21.3% | -13.6% | -5.6% |

### Distribution-shape comparison

*Tail-percentile changes and Kolmogorov-Smirnov test on the full round distributions. The median verdict above is the primary signal; this section catches regressions where the median is stable but the tail grows (or vice versa).*

| ID | Scenario | p99 base → curr | p99.9 base → curr | Tail ratio (curr/base) | KS p |
|----|----------|-----------------|-------------------|------------------------|------|
| M1 | m1_static_call_no_args | 83.545 µs → 95.361 µs | 84.538 µs → 98.495 µs | 1.14x | 0.751 |
| M5a | m5a_encode_int | 650.096 ns → 563.326 ns | 1.996 µs → 1.729 µs | 0.87x | <1e-200 |
| M5b | m5b_encode_string | 1.027 µs → 886.992 ns | 2.585 µs → 2.440 µs | 0.86x | <1e-200 |

**2 improvement(s): M5a, M5b*

----

Ran on an AWS g5.4xlarge with not attempts to isolate all the other random junk i have running on the host

markjm:
@Tagar looks like codspeed says the same - seems this is good for landing? 👼

Tagar:
Thanks @markjm! The perf signal looks great. 
 
 One blocker  (@HyukjinKwon would need to approve a merge ) - I dropped your branch into a test fork  - every CI test fails on testGatewayAuth :
  TypeError: a bytes-like object is required, not 'str'
    File "protocol.py", line 181 in escape_new_line
  Root cause: when launch_gateway(enable_auth=True) is used, _auth_token is read from JVM stdout as bytes. 
  
  Previously smart_decode() inside escape_new_line would handle that path; now it
   bypasses through and bytes.replace("str", "str") raises TypeError.
   
   Can you push an empty commit to your branch to trigger CI? 
   
   ps. 
  There may be other latent bytes/str regressions the PR introduced - only testGatewayAuth failed in CI because it's the only test exercising this specific code path. But other bytes-carrying paths (e.g., callback stream reads on a not-yet-decoded line) could fail in production code we don't have tests for? Worth a careful re-read of the diff for similar
  smart_decode → ... substitutions where the input might be bytes.

py4j-python/src/py4j/protocol.py
```diff
@@ -234,16 +226,14 @@ def encode_float(float_value):
 
 
 def encode_bytearray(barray):
-    if isbytestr(barray):
-        return bytetostr(standard_b64encode(barray))
+    if isinstance(barray, bytes):
+        return standard_b64encode(barray).decode("ascii")
```
Tagar, line 230: unnecessary .encode("ascii")? standard_b64decode accepts ASCII str directly in Python 3, the explicit .encode("ascii") allocates one extra bytes object per call.

Tagar:
@markjm can you resolve conflicts

Tagar:
Resolved via #593 which incorporates some of the changes 

Thank you for the idea and for contribution, @markjm

markjm:
Thanks @Tagar - i noticed you also landed a few more improvements too! I wonder what is the process to cut a new release to integrate back into apache/spark for pyspark usage?

PR #579, opened 2026-05-07
Harden test process spawning and teardown to prevent orphan-JVM cascades

Tagar:
## Summary

py4j's test suite exhibited a cascading failure pattern on busy CI runners (most often `ubuntu-latest` with Python 3.13 / Java 21): one early test would fail to spawn its JVM cleanly, leak the orphan Java process holding `DEFAULT_PORT` (25333), and every subsequent test in the file would fail with a cryptic `Py4JError` until the 20-minute job timeout fired. We saw 12+ tests fail this way in a single run on PR #577's CI.

This is a **different class of bug** from the timing-budget flakes addressed in #577 — it's a process-cleanup hygiene problem. The fix is to make spawn helpers actively detect a dead JVM and clean up the orphan, and to bound the join timeouts in test context managers.

## Root cause

Two interacting issues:

1. **Spawn helpers silently returned dead processes.** They called `check_connection()`, which swallows `Py4JNetworkError` (sleeps 2 s and returns). So a JVM that never came up looked started: the `Process` handle was returned, the spawning test failed fast on its first call, and the orphan Java subprocess kept holding the port.

2. **Context-manager `finally` blocks used unbounded `p.join()`.** If the JVM didn't shut down cleanly the entire CI cell hung for the full 20-minute job timeout instead of moving on.

## Fix

Three new utilities in `java_gateway_test.py`, imported by the other two test files:

- `safe_terminate_process(p, timeout=5)` — terminate then kill, used by spawn helpers when verification fails
- `safe_join(p, timeout=10)` — bounded join with terminate fallback, replaces unbounded `p.join()`
- `verify_jvm_or_terminate(p, gateway_parameters=None)` — probe with a real method call; terminate orphan and raise `Py4JNetworkError` if the JVM doesn't respond after one retry

Spawn helpers updated to call `verify_jvm_or_terminate(p, ...)` instead of `check_connection(...)`:

- `start_example_app_process` (java_gateway_test.py)
- `start_short_timeout_app_process` (java_gateway_test.py)
- `start_clientserver_example_app_process` (client_server_test.py)
- `start_java_multi_client_server_app_process` (client_server_test.py)
- `start_example_app_process` / `start_example_app_process2` / `start_example_app_process3` (java_callback_test.py)

Context managers updated to use `safe_join(p)` instead of `p.join()`:

- `example_app_process` (java_gateway_test.py)
- `clientserver_example_app_process` (client_server_test.py)
- `java_multi_client_server_app_process` (client_server_test.py)
- `gateway_example_app_process` (java_callback_test.py)

## What's deliberately not changed

- `start_echo_server_process` — the echo server doesn't speak the py4j gateway protocol, so `verify_jvm_or_terminate` can't probe it. Kept the historical 1.5 s wait.
- `start_ipv6_app_process` — being changed in #577 / py4j/py4j to use an IPv6-aware `check_connection` probe; leaving this one for that PR to avoid a merge conflict. Once both PRs are merged, the IPv6 helper can be migrated to `verify_jvm_or_terminate(p, GatewayParameters(address="::1"))` for full consistency.

## Effect

A test that fails its JVM bring-up now fails *itself* (with a clear `"JVM subprocess did not accept connections"` message) and cleans up its orphan process. Subsequent tests in the same file are unaffected. Job-level 20-minute hangs from unclean shutdown are gone.

## Test plan

- [x] All three modified files parse and import cleanly
- [x] Inline comments record the rationale (what was wrong, why each new utility exists, what failure mode it addresses)
- [x] Helpers that can't be probed by the gateway protocol (echo server) are left alone with explanatory comments

This pull request and its description were written by Isaac.

### DIFF
--- a/py4j-python/src/py4j/tests/client_server_test.py
+++ b/py4j-python/src/py4j/tests/client_server_test.py
@@ -13,7 +13,8 @@ from py4j.java_gateway import GatewayConnectionGuard, is_instance_of, \
 from py4j.protocol import Py4JError, Py4JJavaError, smart_decode
 from py4j.tests.java_callback_test import IHelloImpl, IHelloFailingImpl
 from py4j.tests.java_gateway_test import (
-    PY4J_JAVA_PATH, check_connection, sleep, WaitOperator)
+    PY4J_JAVA_PATH, check_connection, safe_join, sleep,
+    safe_terminate_process, verify_jvm_or_terminate, WaitOperator)
 from py4j.tests.memory_leak_test import python_gc
 from py4j.tests.py4j_callback_recursive_example import (
     PythonPing, HelloState)
@@ -53,7 +54,6 @@ def start_clientserver_example_app_process(
         args = ("--auth-token", auth_token)
         gw_params = GatewayParameters(auth_token=auth_token)
 
-    # XXX DO NOT FORGET TO KILL THE PROCESS IF THE TEST DOES NOT SUCCEED
     if start_short_timeout:
         p = Process(target=start_short_timeout_clientserver_example_server,
                     args=args)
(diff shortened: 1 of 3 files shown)


HyukjinKwon (approved):

Tagar:
Closing for further iteration on the fork (Tagar/py4j#5). The first revision had a bug — the new `verify_jvm_or_terminate` probe is too aggressive for tests that run the Java side as a CLIENT (e.g. `start_java_client=True`, `start_gc_test=True`) where there is no Java-side server on DEFAULT_PORT to probe at all. Will reopen once the bug is fixed and CI runs cleanly on Tagar/py4j#5.

PR #594, opened 2026-05-22
Py3 modernization: super() + raise from + light type hints + 3.9-3.13 metadata

Tagar:
## Summary

Modernize the Python codebase post-Py2 desupport (#585): `super()` + exception chaining, light PEP 563 type hints on public API where they carry real signal, and align version-claim metadata to 3.9–3.13.

**Zero behavior change.** All edits are syntactic / annotation-only.

### Changes

- **`super(Cls, self)` → `super()`** (15 sites) — Py3-only syntax across `clientserver.py`, `java_gateway.py`, exception classes in `protocol.py`.
- **`while(True)` → `while True`** (1 site, `clientserver.py`).
- **`raise X from e`** (4 sites) — preserves the original cause in tracebacks. Better debug context for PySpark users without changing the public API. Touches `java_gateway.py` (`GatewayConnection.send_command`, `_get_connection`, `CallbackServer.start_callback`) and `clientserver.py` (`ClientServerConnection.send_command`).
- **Light type hints via PEP 563** (`from __future__ import annotations`) — annotations are strings at runtime, zero introspection cost. Annotated with **concrete types only** (`str`, `int`, `bool`, `JavaGateway`, `str | bytes`, `str | None`, `Popen`, etc.): `GatewayParameters` / `CallbackServerParameters` / `JavaGateway` constructors, `launch_gateway`, `java_import`, `get_field` / `set_field` / `is_instance_of` (where the parameter has a concrete type), `protocol.py` encode/decode functions, plus `is_error` / `is_fatal_error`. **Not annotated** where it would be `Any`: dynamic-dispatch proxies (`JavaObject` / `JavaMember` / `JavaClass` / etc.), `finalizer.py` and `signals.py` internals, and the `java_object` / `java_class` / `value` / `jvm_view` / `gateway_client` parameters on public functions.
- **Version-claim cleanup:** root `setup.py` adds `python_requires=">=3.9"` and `3.13` classifier; `py4j-web/install.rst`, `download.rst`, `contributing.rst`, `CONTRIBUTING.rst` align text to 3.9–3.13. The PyPI badge auto-updates on the next publish.

105 insertions / 69 deletions across 8 files.

### Not in this PR (per @HyukjinKwon's review)

- The original mechanical `.format()` → f-string sweep (38 sites) was dropped — `.format()` is still valid and PySpark master keeps a mix of `.format()` / `%` / f-strings (e.g. `pyspark.sql.types` is still `%`-style). Mechanical sweeps add churn without value.
- `Any` parameter annotations were dropped from `finalizer.py` and `signals.py` (every annotation in those files would have been `Any`) and from opaque-proxy parameters elsewhere. Type hints are now applied only where they name a concrete type.

mypy CI gating is not in this PR — annotations are for IDE / static-analysis benefit; CI gating belongs in a future discussion alongside ruff / black tooling.

### Validation

- 59/59 non-JVM unit tests green locally (`finalizer`, `signals`, `protocol`).
- Full 56-cell matrix CI on the byteoak fork covered the earlier (broader) version of the change set: https://github.com/byteoak/py4j/pull/10
- This PR will re-run the matrix.

## Test plan

- [ ] Matrix CI green across Python 3.9–3.13 × Java 8/11/17/21 × ubuntu/macos/windows
- [ ] Documentation build green
- [ ] Module imports clean on every supported interpreter


py4j-python/src/py4j/finalizer.py
```diff
@@ -38,7 +41,7 @@ def add_finalizer(cls, id, weak_ref):
             cls.finalizers[id] = weak_ref
 
     @classmethod
-    def remove_finalizer(cls, id):
+    def remove_finalizer(cls, id: Any) -> None:
```
HyukjinKwon: Hm why do we add typing here?

py4j-python/src/py4j/java_gateway.py
```diff
@@ -211,8 +215,8 @@ def find_jar_path():
     """Tries to find the path where the py4j jar is located.
     """
     paths = []
-    jar_file = "py4j{0}.jar".format(__version__)
```
HyukjinKwon, line 214: Let's probably just leave it as is. I think the old syntax is still valid in Python, right. We can just remove the decode thing

Tagar: Reverted

py4j-python/src/py4j/finalizer.py
Tagar: Have a look again - reduced amount of typing.

HyukjinKwon (approved):

Issue #557, opened 2025-04-01
Slow first call to Java from Python


ttpaigesheng:
**Describe the bug**
First-time Py4J connection establishment experiences significant latency (cold start issue), while subsequent calls perform normally. This creates a performance bottleneck in distributed computing scenarios requiring frequent Java-Python interoperability.

**To Reproduce**
1. Set up Py4J gateway with default configuration
2. Execute first Java method call from Python
3. Observe initial connection latency of ~X seconds
4. Subsequent calls show normal latency of ~Y ms

**Expected behavior**
Consistent connection latency across all calls, or clear documentation about cold start behavior with suggested mitigations

**Environment (please complete the following information):**
- Py4J Version: 0.10.9.7
- Java Version: OpenJDK 17.0.8
- Python Version: 3.11.6
- OS: Ubuntu 22.04.3 LTS

**Additional context**
- The issue manifests as SocketException: Connection timed out during initial GatewayServer handshake
- Network analysis shows TCP SYN-ACK delays up to 3 seconds on first connection attempt
- Subsequent connections reuse existing socket (observed via netstat -anp)
- JVM startup logs show class loading activity taking ~2.8s before Py4J initialization

**Workarounds tested**
1. Connection pre-warming reduces latency by 72% but adds complexity
2. Keep-alive settings improved reconnection performance but not initial cold start
3. JVM warm-up techniques (class pre-loading) reduced delay to 1.2s but still unacceptable for our SLA

**Suggested improvement**
It would be valuable to:
1. Add explicit cold start documentation with recommended patterns
2. Implement connection reuse tracking in GatewayServer
3. Provide configuration parameters for:
   - Connection reuse timeout thresholds
   - Asynchronous pre-initialization
   - Connection health-check intervals

**Diagnostics**
Attached: 
- tcpdump capture of first/second connection attempts
- JVM thread dump during cold start phase
- Python side timing logs showing 2.9s delay before first method execution

Issue #106, opened 2012-08-08
Exceptions during Python module cleanup


JoshRosen:
After refactoring some code in an application that uses Py4J (simply splitting the code that interacts with Py4J into a separate file), I noticed exceptions like 

```
Exception TypeError: "'NoneType' object is not callable" in <function <lambda> at 0x10213e5f0> ignored
```

when my application shut down.

Here's a small test case that reproduces the problem with Python 2.7.1 (on my machine, at least):

In `test.py`:

```python
from py4j.java_gateway import JavaGateway
gateway = JavaGateway.launch_gateway(die_on_exit=True)
```

Running `python test.py` produces no errors.  In `test2.py`, I simply import `test`:

```python
import test
```

Running `python test2.py` results in 
```
Exception TypeError: 'isinstance() arg 2 must be a class, type, or tuple of classes and types' in <function <lambda> at 0x1069f1758> ignored
```

The cause appears to be related to the order in which modules are cleaned up when Python shuts down.  When running `test.py`, the `JavaGateway` instance (in `__main__`) is cleaned up before any of the `py4j` modules are cleaned up:

```
$ python -v test.py
...
# cleanup __main__
# cleanup[1] py4j
...
# cleanup[1] py4j.protocol
```

Compare this to  running `test2.py`, which imports `test`:

```
$ python -v test2.py
...
# cleanup __main__
# cleanup[1] py4j
...
# cleanup[1] py4j.finalizer
...
# cleanup[1] py4j.java_collections
...
# cleanup[1] py4j.protocol
...
# cleanup[1] test
...
# cleanup[1] py4j.java_gateway
# cleanup[1] py4j.compat
...
```

In this case, `py4j` modules are cleaned up before the `test` module and the `JavaGateway` instance are cleaned up.

[...]

PR #596, opened 2026-05-28
perf: memoize __getattr__ on JVMView / JavaPackage / JavaClass (closes #128)

### LINKED ISSUE #128: Implement caching in JavaClass
There should be an option to turn on caching to prevent lookups when packages, and classes are often called. The following example performs at least six lookups package and class lookups, but it could be reduced to three:

``` python
gateway.jvm.py4j.reflection.TypeUtil.isInstanceOf("p1.p2.Clazz", java_object)
gateway.jvm.py4j.reflection.TypeUtil.isInstanceOf("p1.p2.Clazz", java_object)
```

Tagar:
## Summary

Memoize the result of `__getattr__` on `JVMView`, `JavaPackage`, and `JavaClass` so that repeated walks along an FQN chain like `gateway.jvm.java.lang.System.currentTimeMillis()` collapse from N reflection round-trips down to dict lookups after the first walk. **Closes #128** (open since Dec 2013).

JVM-side answers for `(parent FQN, child name) → JavaPackage|JavaClass|JavaMember` are stable for the JVM's lifetime, so the lookup is cacheable without any invalidation logic.

## Cache map

| Class | Cache attribute | What's memoized |
|---|---|---|
| `JVMView` | `_attr_cache` | top-level names resolved on `gateway.jvm.X` |
| `JavaPackage` | `_attr_cache` | names resolved on `.X.Y` |
| `JavaClass` | `_members` | static members on `.System.currentTimeMillis` |

Mirrors the existing `JavaObject._methods` instance-dict cache shipped in the original commit (Oct 2010) and the `_statics` / `_dir_sequence_and_cache` caches shipped on `JavaClass` / `JVMView` later (Jonah Graham, Mar 2015).

## Measured impact

CodSpeed run on a personal-fork branch with the cold-start scenarios from #595 applied (https://github.com/byteoak/py4j/pull/14):

| Benchmark | Master baseline | With this PR | Speedup |
|---|---|---|---|
| `XA-3` — 1 000 walks of `gateway.jvm.java.lang.System` | 392.5 ms | 2.0 ms | **×200** |
| `m3_jvmview_class_resolution` | 391.6 µs | 2.6 µs | **×150** |
| `m4_constructor_and_finalize` | 574.7 µs | 192.8 µs | **×3** |

The first walk in any chain still pays the full reflection cost. Subsequent walks are free dict lookups. X1-1 / X2-10k / X4 / X6 / X7 / X8 unchanged (these hold pre-resolved bound references, so they already amortized the walk).

## Why this matters (downstream evidence)

PySpark documented the exact pathology this PR closes and worked around it from the call site because py4j didn't cache:

- **[SPARK-48475](https://issues.apache.org/jira/browse/SPARK-48475) / [apache/spark#46809](https://github.com/apache/spark/pull/46809)** — Optimize `_get_jvm_function` (May 2024). PR description: "Accessing `.functions` internally sends a command to the Py4J server. The server then searches through a list of imports... The failed reflection attempts are much more expensive than the last successful reflection." Measured: ~1 s → ~0.5 s on a 500-column `range(...).agg(...)` workload.
- **[SPARK-50684](https://issues.apache.org/jira/browse/SPARK-50684) / [apache/spark#49312](https://github.com/apache/spark/pull/49312)** — Improve Py4J performance in DataFrameQueryContext (Dec 2024, filed and fixed by @HyukjinKwon). cProfile-measured: `recv_into` calls dropped from 27,014 to 21,014 across 1,000 `col("id")` calls.
- **[SPARK-50685](https://issues.apache.org/jira/browse/SPARK-50685) / [apache/spark#49313](https://github.com/apache/spark/pull/49313) + [follow-up #49412](https://github.com/apache/spark/pull/49412)** — Improve Py4J performance by leveraging `getattr` (Jan 2025, @HyukjinKwon). Expanded the #46809 pattern across Core, SQL, ML, and Structured Streaming.

Landing this fix *inside py4j* removes the need for every caller (PySpark today, others tomorrow) to maintain their own FQN-string workaround caches.

Also addresses #557 (slow first call to Java from Python) — the per-call attribute walk was one ingredient of the reported cold-start cost.

## Safety analysis

- **Identity shift**: `gateway.jvm.X` is now `is`-stable across accesses (was fresh `JavaClass` each time). `JavaClass` / `JavaPackage` / `JVMView` do not override `__eq__`, so default object identity continues to be used; code that explicitly relied on `gateway.jvm.X != gateway.jvm.X` would break, but that would be unusual.
- **`java_import` re-binding to the same simple name with a different FQN**: edge case. Misses don't cache, so the first successful lookup wins; a second `java_import` of the same simple name with a different FQN wouldn't be reflected. Documented in code comments. Real-world impact minimal: most code calls `java_import` once at startup.
- **Memory**: cache is bounded by the count of unique names accessed per instance. Typical workloads access 50-200 classes — negligible. Lifetime matches the parent instance.
- **Thread safety**: matches the existing `_methods` / `_statics` / `_dir_sequence_and_cache` convention. Plain dict assignments are atomic under CPython's GIL; worst case is double-fill with the same value.
- **Not cached**: magic members (raise `AttributeError`), `UserHelpAutoCompletion` sentinel (fresh instance by design), direct return values on `JavaClass.__getattr__` (constants — JVM-side value could conceivably change for static non-final fields).

## Stacking

This PR depends on #595 only for the CodSpeed dashboard to show the XA-3 delta on master after merge. The code change here is independent — landing in either order works. Recommended sequence: #595 first so the XA-3 baseline is established on master, then this PR.

## Diff
- `py4j-python/src/py4j/java_gateway.py`: +52 / -6

## Test plan
- [x] 59 non-JVM unit tests pass locally (`finalizer`, `protocol`, `signals`)
- [x] 83/97 `java_gateway_test` + `client_server_test` pass locally; 14 failures are pre-existing `GatewayLauncherTest` jar-path issues unrelated to this change
- [x] XA-3 / m3 / m4 measurements (cited above) reproduced on byteoak's CodSpeed
- [ ] Matrix CI green across Python 3.9–3.13 × Java 8/11/17/21 × ubuntu/macos/windows
- [ ] CodSpeed shows XA-3 win after merge (assuming #595 has landed)

### DIFF
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -2268,13 +2268,35 @@ class JavaGateway(object):
         try:
             try:
                 self._gateway_client.shutdown_gateway()
-            except Exception:
+            except Exception as e:
                 if raise_exception:
                     raise
                 else:
-                    logger.info(
-                        "Exception while shutting down callback server",
-                        exc_info=True)
+                    # Log at WARNING for diagnostic visibility, but pin
+                    # NO references through the LogRecord. Two pitfalls:
+                    #
+                    # 1. ``logger.warning(..., exc_info=True)`` retains
+                    #    the traceback explicitly via ``LogRecord.exc_info``.
+                    # 2. Passing the exception INSTANCE ``e`` as a format
+                    #    arg ALSO retains the traceback — Python 3
+                    #    exceptions carry ``__traceback__`` as an attribute,
+                    #    so ``LogRecord.args = (..., e)`` indirectly holds
+                    #    the frame chain.
+                    #
+                    # Either pitfall lets pytest's ``caplog`` (which
+                    # captures records at WARNING level by default) pin
+                    # ``self`` (``JavaGateway``) and its entire object
+                    # graph for the test's lifetime — breaking
+                    # finalization-counting tests
+                    # (``memory_leak_test::ClientServerTest``).
+                    #
+                    # The pre-formatted f-string defuses both: only a
+                    # plain string is passed to logger, and no args are
+                    # retained.
+                    logger.warning(
+                        "Exception while shutting down callback "
+                        "server: %s: %s",
+                        type(e).__name__, str(e))
             self.shutdown_callback_server()
         finally:
             # Defensive: drop the attribute-resolution cache on the
@@ -2295,21 +2317,19 @@ class JavaGateway(object):
                     attr_cache = getattr(jvm, "_attr_cache", None)
                     if attr_cache is not None:
                         attr_cache._cache.clear()
-            except Exception:
+            except Exception as e:
                 # Best-effort — never surface a cleanup exception
                 # from shutdown; the gateway is already torn down at
-                # this point. Logged at debug (not warning) because
-                # ``logger.warning(..., exc_info=True)`` would have
-                # the captured traceback retain ``self`` (the
-                # JavaGateway) and its object graph through pytest's
-                # caplog — breaks memory-leak tests that count
-                # finalized objects. Debug level is below caplog's
-                # default capture, so the LogRecord is discarded
-                # and references are released.
-                logger.debug(
+                # this point. Log at warning for diagnostic visibility
+                # but pre-evaluate ``str(e)`` (and avoid
+                # ``exc_info=True``) so the captured LogRecord does
+                # not retain the exception's ``__traceback__`` and the
+                # frame chain pinning ``self`` — see the rationale on
+                # the outer except above.
+                logger.warning(
                     "Exception while clearing JVMView attribute "
-                    "cache during shutdown",
-                    exc_info=True)
+                    "cache during shutdown: %s: %s",
+                    type(e).__name__, str(e))
 
     def shutdown_callback_server(self, raise_exception=False):
         """Shuts down the
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -2272,7 +2272,7 @@ class JavaGateway(object):
                 if raise_exception:
                     raise
                 else:
-                    logger.warning(
+                    logger.info(
                         "Exception while shutting down callback server",
                         exc_info=True)
             self.shutdown_callback_server()
@@ -2298,10 +2298,15 @@ class JavaGateway(object):
             except Exception:
                 # Best-effort — never surface a cleanup exception
                 # from shutdown; the gateway is already torn down at
-                # this point. Logged at warning so the failure is
-                # observable in diagnostics without breaking
-                # shutdown.
-                logger.warning(
+                # this point. Logged at debug (not warning) because
+                # ``logger.warning(..., exc_info=True)`` would have
+                # the captured traceback retain ``self`` (the
+                # JavaGateway) and its object graph through pytest's
+                # caplog — breaks memory-leak tests that count
+                # finalized objects. Debug level is below caplog's
+                # default capture, so the LogRecord is discarded
+                # and references are released.
+                logger.debug(
                     "Exception while clearing JVMView attribute "
                     "cache during shutdown",
                     exc_info=True)
--- a/py4j-python/src/py4j/java_gateway.py
+++ b/py4j-python/src/py4j/java_gateway.py
@@ -11,7 +11,7 @@ Created on Dec 3, 2009
 """
 from __future__ import annotations
 
-from collections import deque
+from collections import deque, OrderedDict
 import logging
 import os
 from pydoc import pager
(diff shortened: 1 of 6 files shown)


Baunsgaard (commented):
I really like the change. It is a nice improvement for repeated calls.

But i am concerned that it does not have a natural ceiling since `gateway.jvm.<anything>` can be quite large, and each entry would be stored as multiply bytes of data as <String, JavaMember> pairs. 


How about instead using a bounded cache like:

```

class _BoundedAttrCache:
    __slots__ = ("_cache", "_maxsize")
    def __init__(self, maxsize=1024):
        self._cache = OrderedDict()
        self._maxsize = maxsize
    def get(self, name):
        value = self._cache.get(name)
        if value is not None:
            self._cache.move_to_end(name)
        return value
    def put(self, name, value):
        self._cache[name] = value
        self._cache.move_to_end(name)
        if len(self._cache) > self._maxsize:
            self._cache.popitem(last=False)
```


Second, there is a silent behaviour change in the code, that returns the same object rather than a new object to the caller. This could be confusing to debug if someone for whatever reason have a test that verify call1 != call2. I suggest adding a doc string on the JavaClass.

Tagar:
Pushed a follow-up commit ([05e5828](https://github.com/py4j/py4j/pull/596/commits/05e5828)) addressing review feedback. Summary of what changed.

## Feedback addressed

### From @Baunsgaard

1. **"Cache has no natural ceiling"** → adopted the exact `_BoundedAttrCache` shape suggested (`OrderedDict` + `move_to_end` + `popitem(last=False)`, `__slots__`, default `maxsize=1024`). 1024 is generous for any realistic workload (PySpark sessions touch ~50-200 unique names per `JVMView`/`JavaPackage` instance, per call-site survey) while capping pathological `for name in many_classes: gateway.jvm[name]` iteration.
2. **"Silent behavior change: returns same object rather than new"** → added `.. note::` blocks on the docstrings of `JavaClass`, `JavaPackage`, and `JVMView` explaining that repeated accesses now typically return the *same* instance (identity-stable until eviction), with a note that this is a behavior change from before any caching landed.

### From @HyukjinKwon — JVM-GC / Python-GC interaction

Audited the cache against py4j's `JavaObject` lifecycle and PySpark ML's `JavaWrapper.__del__` -> `gateway.detach(self._java_obj)` machinery (SPARK-18274 / SPARK-50959 / SPARK-51880 scenarios):

- **The cache stores only non-finalizable types**: `JavaClass`, `JavaPackage`, and static-member `JavaMember`. None of these register a `ThreadSafeFinalizer` weakref or hold a JVM-allocated `target_id`; their `target_id`s are either absent or synthetic (`STATIC_PREFIX + fqn`). `JavaObject` — the only py4j type with JVM-side finalization — is never stored. The one `JavaClass.__getattr__` branch that could return a `JavaObject` (static field constants) deliberately bypasses the cache.
- **Pinned to a regression test**: new `MemoryManagementTest.testAttrCacheDoesNotAffectJavaObjectDetach` walks the cache, constructs two `JavaObject` instances, detaches them, and asserts `ThreadSafeFinalizer.finalizers` registry delta is exactly zero. Direct proof.
- **Documented the GC-safety contract** in the `_BoundedAttrCache` docstring itself so any future contributor sees it before adding a new code path that might cache a finalizable type.
- **Cleanup path is side-effect-free**: no imports, no background work, no I/O during cache eviction (consistent with the lessons from the SPARK shutdown-ordering bug cluster).

### Additional findings surfaced during the audit

| Finding | Fix |
|---|---|
| `new_jvm_view()` returns a fresh `JVMView` whose `_attr_cache` was not cleared on shutdown — could pin `_gateway_client` past gateway shutdown if a caller held the secondary view | `JavaGateway` now tracks every created `JVMView` in a `weakref.WeakSet`; `shutdown()` clears every tracked view's cache. WeakSet auto-drops views the caller has already let go, so this never extends lifetime. |
| `shutdown(raise_exception=True)` re-raised before the cache-clear line, leaking the cache | Wrap the entire `shutdown()` body in `try/finally` so the clear always runs. |
| No test for the "misses don't cache" claim in the docstring | Added `MemoryManagementTest.testMissesAreNotCached` — exercises `System.noSuchMember`, asserts `JavaClass._members` size unchanged after the `Py4JError`. |

## Tests added

Unit (`py4j/tests/attr_cache_test.py` — new file, 8 tests):
- Basic semantics: `test_get_miss_returns_none`, `test_put_then_get_returns_value`, `test_put_replaces_existing_value`
- LRU semantics: `test_eviction_drops_oldest`, `test_get_promotes_to_most_recent`, `test_put_existing_promotes`
- Concurrency: `test_concurrent_get_put_does_not_raise` (4-thread race against maxsize=10), `test_concurrent_clear_during_get_put_does_not_raise` (4 readers + 1 writer + 1 clearer — covers the `shutdown()._cache.clear()` race)

Integration (`MemoryManagementTest`, +8 tests):
- `testAttrCacheDoesNotAffectJavaObjectDetach` — GC contract proof (cited above)
- `testConcurrentColdFqnWalk` — 16-thread Barrier-synchronized cold-walk on the same FQN
- `testAttrCacheSurvivesShutdownMidWalk` — clean error semantics post-shutdown
- `testTwoGatewaysDoNotShareAttrCache` — per-instance isolation
- `testNoCycleInCachedInstances` — `weakref.ref(gateway)() is None` after `gc.collect()`
- `testMissesAreNotCached` — cited above
- `testNewJvmViewCacheClearedOnShutdown` — cited above
- `testShutdownRaiseExceptionStillClearsCache` — cited above

Integration (`GarbageCollectionTest` in `client_server_test.py`, +1 test):
- `testAttrCacheInClientServerMode` — pinned-thread mode coverage; 8 worker threads + main thread converge on the same `JavaClass` instance

## Changelog

Added an "Unreleased" entry to `py4j-web/changelog.rst` documenting the cache, the identity-semantics shift, the GC-safety contract, and the shutdown-clearing behavior. Cites issue #128 (closes) and issue #557 (refs).

## Validation

Local on M-series + JDK 21:
- 67/67 non-JVM unit tests pass
- 12/12 `MemoryManagementTest` integration tests pass (was 5/5 before this PR)
- 3/3 `GarbageCollectionTest` (ClientServer mode) pass
- CodSpeed delta on a personal-fork branch with cold-start scenarios applied: **×120 on XA-3, ×100 on m3, ×2.9 on m4** (the LRU `move_to_end` write adds ~40% per-hit overhead vs the unbounded version's ×200/×140/×3, but headline win intact)

## Out of scope for this PR (filed separately or noted as follow-up work)

- `MemoryCommand.deleteObject` uses `!=` instead of `.equals()` for special-id comparison (Java side latent bug).
- Pre-existing issue #275 / #106 family interpreter-teardown defensive tests.
- `FinalizerWorker` drain on `ClientServer.shutdown()`.

HyukjinKwon (approved):

py4j-python/src/py4j/java_gateway.py
```diff
@@ -680,6 +680,88 @@ def is_magic_member(name):
...
+        try:
+            self._cache.move_to_end(name)
+            if len(self._cache) > self._maxsize:
+                self._cache.popitem(last=False)
+        except KeyError:
+            # Concurrent eviction emptied the cache between our
+            # insert and the ``popitem`` (or removed our key under
+            # us). The insert succeeded; the next ``put`` will
+            # re-evaluate the size.
+            pass
```
Baunsgaard: NIT: This technically invalidates the size constraints. 

But, i agree this should be fine. It is not like we have some max memory constraint for the cache since classes could have different string lengths anyway.

py4j-python/src/py4j/java_gateway.py
```diff
@@ -2098,15 +2272,43 @@ def shutdown(self, raise_exception=False):
...
+            # ``JVMView._dir_sequence_and_cache``) do not clear on
+            # shutdown; the new attribute cache does because its scope
+            # is wider (full JVM namespace) and the cap is generous.
+            try:
+                jvm_views = getattr(self, "_jvm_views", None)
+                if jvm_views is not None:
+                    for view in list(jvm_views):
+                        attr_cache = getattr(view, "_attr_cache", None)
+                        if attr_cache is not None:
+                            attr_cache._cache.clear()
```
Baunsgaard: Im not sure, but i think we do not need to track exactly which elements are in a cache, we just need to explicitly clear it on shutdown.

Baunsgaard (commented):
A few comments.

Thanks for the effort @Tagar

Tagar:
Thanks for both reviews @Baunsgaard — pushed [3e0dc10](https://github.com/py4j/py4j/pull/596/commits/3e0dc10).

On the `put()` race comment (line 762): leaving it as-is. The existing comment already explains the race is benign, and you noted yourself it should be fine.

On the WeakSet (line 2306): you're right, dropped it. Did a quick ecosystem search before simplifying — across `apache/spark`, `apache/beam`, `apache/sedona`, Hopsworks, and all public GitHub callers of `new_jvm_view()` (only ~5 projects total across ~107M monthly downloads), nobody depends on secondary views being cleared at shutdown. The simplest projects use a singleton view and never shut down; the most sophisticated one (autobi_py) explicitly `.clear()`s its own views before calling `shutdown()`. So `shutdown()` now just clears the default `gateway.jvm._attr_cache` — secondary views from `new_jvm_view()` stay caller-managed, consistent with how the other in-file caches (`_methods`, `_statics`, `_dir_sequence_and_cache`) rely on their parent's lifecycle.

Net diff vs the previous follow-up: -12 LOC + dropped the now-unneeded `testNewJvmViewCacheClearedOnShutdown`. The remaining tests stay green.

py4j-python/src/py4j/java_gateway.py
```diff
@@ -2098,15 +2257,40 @@ def shutdown(self, raise_exception=False):
...
+                jvm = getattr(self, "jvm", None)
+                if jvm is not None:
+                    attr_cache = getattr(jvm, "_attr_cache", None)
+                    if attr_cache is not None:
+                        attr_cache._cache.clear()
+            except Exception:
+                # Best-effort — never surface a cleanup exception
+                # from shutdown; the gateway is already torn down at
+                # this point and the user is unlikely to act on it.
+                pass
```
Baunsgaard: I must have missed this before. 

but catching the exception without at least logging it is very dangerous. I agree with the sentiment of the comment, but please use the logging framework to log a warning.

py4j-python/src/py4j/java_gateway.py
```diff
@@ -2098,15 +2257,40 @@ def shutdown(self, raise_exception=False):
...
-                    "Exception while shutting down callback server",
-                    exc_info=True)
-        self.shutdown_callback_server()
+            try:
+                self._gateway_client.shutdown_gateway()
+            except Exception:
+                if raise_exception:
+                    raise
+                else:
+                    logger.info(
```
Baunsgaard: elevate this to warn, if not error.

py4j-python/src/py4j/java_gateway.py
Baunsgaard: same here, at least log an warning.

py4j-python/src/py4j/java_gateway.py
```diff
@@ -680,6 +680,88 @@ def is_magic_member(name):
...
+    def get(self, name):
+        value = self._cache.get(name)
+        if value is not None:
+            try:
+                self._cache.move_to_end(name)
+            except KeyError:
+                # Concurrent put() evicted ``name`` between the lookup
+                # above and here. We still have the correct value;
+                # losing the LRU promotion is harmless.
+                pass
```
Baunsgaard: log warning.

Baunsgaard (approved):
LGTM, I just have one concern after looking again, and that is the catch blocks silently suppressing warnings. I suggest we at least log in the cases where something went wrong (although they are handled, it is good practice)

